PR #2, opened 2018-08-01
Implement packed attestation data validation

aseigler:


abergs:
Nice work! Really thankful for the contributions and that you solved the conversion for COSE_KEY!
🎉

PR #24, opened 2018-08-23
All conformance tests now pass in Server-ServerAuthenticatorAttestati…

aseigler:
…onResponse-Resp-4 except ALG_SIGN_ED25519_EDDSA_SHA256_RAW (check all algorithm boxes).

ASN.1 signature parser now correctly handles long form (> 127 byte) signatures.


abergs:
Great!

PR #33, opened 2018-08-30
Convert fido-net-lib from .NET Core 2.1 to .NET Standard 2.0

ondrokrc:
Proposing changing the fido2-net-lib project type to .NET Standard. It requires little change in code and the lib would be referenceable in all .NET implementations (see also [.NET Standard documentation](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)).


abergs:
@ondrokrc I agree. It should be .NET standard. Will run tests on the PR before merging.

abergs:
Well done, thank you.

PR #35, opened 2018-08-31
Fixed userVerification logic to pass conformance tests

aseigler:
@abergs, please check AssertionOptionsTest in particular and the other related to ensure what I have done is acceptable.  It seems that the conformance test is pushing authenticatorSelection -> userVerification and we were not set up to do anything with that, which was causing lost data resulting in some confusion.

There is some inconsistency and in some cases direct conflict between the conformance tests and the published specification, I am hoping that this will be resolved at some point soon.


aseigler:
No worries.  I have gotten what I needed.  Thanks for answering!

abergs:
@aseigler I thought i'd hold off merging this until you give green light since your working on it. Also there are some conflicts which we need to resolve before merge.

aseigler:
I am not seeing any conflicts? Please do check AssertionOptionsTest in particular and the other related to ensure what I have done is acceptable. It resolves the conformance test but might not be proper. Some of the conformance tests seem to directly oppose the published specs.

abergs:
Yeah it seems all good now (weirdly enough?). I agree with the tests being a bit off spec. Might try to get Ackermans opinion on them. Also we push the settings (from authenticatorSelection) in our controller which will not transfer to any consuming clients. Not feeling 100% sure what's the right thing to do, but will take a look tomorrow when I'm not on my phone..

abergs:
Just realized the controller code doesn't  matter. Code reviews on phones are confusingly hard ;)

abergs:
I say we go with this. I think our best bet to have a well built & maintainable system is to align with the conformance testing tool.

PR #36, opened 2018-09-03
Client JavaScript improvements

ondrokrc:
1. unify POST calling methods (replace _$.post()_ with _fetch()_ which also reduces jQuery dependency for the critical calls + error handling similar to the assertionOptions method)
2. do not show success dialog on registration cancel (check whether confirmed or cancelled)
3. Show registration dialog after successful makeCredentialOptions (similarily to the Login dialog is shown after assertionOptions)


abergs:
Thank you, this was overdue. Client is really not good for demo purposes yet. We should probably add a lot off comments and make it simpler if we want to use it as a reference. I will merge your changes and test afterwards.

PR #47, opened 2018-10-04
Swappable MDS

abergs:
A concept for swappable MDS.

This implementation is "unsafe but easy" by default. The default constructor of Configuration sets the MetdataService to null which is unsafe, but easy to get started.

I'm wondering if this is OK, given that we make it very clear in all examples / documents that you should set the Metadataservice (as shown in the examples in this PR).

I also changed the name of the AppSecrets to follow the standard used by our other variables.

### DIFF
--- a/Fido2Demo/Controller.cs
+++ b/Fido2Demo/Controller.cs
@@ -26,11 +26,12 @@ namespace Fido2Demo
 
         public MyController(IConfiguration config)
         {
-            _lib = new Fido2(new Fido2.Configuration
+            _lib = new Fido2(new Fido2.Configuration()
             {
                 ServerDomain = config["fido2:serverDomain"],
                 ServerName = "Fido2 test",
-                Origin = config["fido2:origin"]
+                Origin = config["fido2:origin"],
+                MetadataService = MDSMetadata.Instance(config["fido2:MDSAccessKey"], config["fido2:MDSCacheDir"])
             });
         }
 
--- a/Fido2Demo/Controller.cs
+++ b/Fido2Demo/Controller.cs
@@ -31,7 +31,7 @@ namespace Fido2Demo
                 ServerDomain = config["fido2:serverDomain"],
                 ServerName = "Fido2 test",
                 Origin = config["fido2:origin"],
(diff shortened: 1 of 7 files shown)


abergs:
This is now published to http://fido2.azurewebsites.net/ but will wait until you can test independently with conformance tests, because I do not receive 100% on the conformance test. The following fails:

Running: FIDO Conformance Tools v0.10.110 (BETA)((BETA) FIDO2 )


Server-ServerAuthenticatorAttestationResponse-Resp-6 Test server processing "packed" SELF(SURROGATE) attestation
F-2 Send ServerAuthenticatorAttestationResponse with SELF "packed" attestation, that contains full attestation, and check that server returns an error

Server-ServerAuthenticatorAttestationResponse-Resp-B Test server processing "android-safetynet" attestation
P-1 Send a valid ServerAuthenticatorAttestationResponse with "android-safetynet" attestation, and check that server succeeds

All tests in Server-ServerAuthenticatorAttestationResponse-Resp-1 Test server processing ServerAuthenticatorAttestationResponse structure
The only passing is P-1.

I might have something wrong in the configuration on the server if these passes for you locally?

aseigler:
The "Self contains a full" test will only pass if you have downloaded the metadata from the conformance tool into the cachdir\Custom directory on the server. The other ones should be clean, with EdDSA support unchecked.

abergs:
@aseigler Ok. It seems that android-safetynet fails against the remote server but works locally which is weird. What does your conformance tests say?

aseigler:
Safetynet has a peculiar timestamp check that none of the others has. I bet that the interpretation server time vs. the client time is the issue. Will run tests when I get a chance tonight.

abergs:
Aha, great!

aseigler:
Run against https://fido2.azurewebsites.net with all algorithms except EdDSA.

image

abergs:
@aseigler Yes, I moved it out from the class in order to not force use of UserSecrets/EnvConfig upon the developer. If the developer is using some other way to store config/secrets they should be able to use what they already have - I also think it makes the API more transparent, e.g. you can discover the parameters needed while you are in the IDE, in case you skipped over the readme.

PR #51, opened 2018-10-05
New conformance tool requirements for UP/UV and some metadata defense

aseigler:


abergs:
Published to live site.

PR #61, opened 2018-12-01
General project structure overhaul

MeikTranel:
Hey @abergs,
this PR includes some of the changes we discussed in https://github.com/MeikTranel/MigrationSupport/issues/1

Next up for me and this project would be adding a build script so that everyone has that first checkout, build and test experience. If that's done a CICD pipeline can just run that script.


MeikTranel:
For the other stuff mentioned I'd first create some issues in this repo for better tracking. I want to get as much people on multi-targeted netstandard as possible. So this PR is done on my part. Merge at your will.

Issue #32, opened 2018-08-30
SHA512 signatures sometimes fail to verify when they should successfully verify (false negative)


aseigler:
SigFromEcDsaSig() attempts to parse an ASN.1 encoded pair of type/length/value triplets into a byte array signature that can be used to verify the integrity of another byte array.  SHA256 and SHA384 result in hashes with length less than 128 bytes, which are well understood and appear to be reliable.  SHA512 hashes seem to be at or just above 128 bytes, which changes how data lengths are presented in ASN.1 encoding format.  I believe some code error with regard to this boundary is the root cause of the occasional false negative data verification result, but I have not been able to pin down exactly what the difference is in a failing vs passing case.

aseigler:
This may no longer be a problem, I have not seen it since that day.

abergs:
Let's close until we get a repro.

aseigler:
https://github.com/dotnet/corefx/issues/33172

herrjemand:
@aseigler @abergs Remove leading 00 from the coefficients when verifying signature. .NET uses RAW ANSI (32 byte R + 32 byte S) and does not support DER signatures with padded coefficients.

aseigler:
I understand where you are coming from, but the problem is, it is actually the other way around in this case with SHA512.  The R value is coming out **without** a leading \00 and the .NET libraries are apparently expecting a fixed sized unsigned big-endian value for R & S, at 66 bytes each for SHA512, whereas OpenSSL is perfectly happy with a 65 byte R + 66 byte S.  I have verified this works, but am struggling to find documentation supporting this logic.

herrjemand:
The signature is a concatenation of two coefficients that are of the size 0 to N, where N is the size order of the curve(P521 -> 521 bits). So coefficient size is (521 * 2 / 8 / 2) or 65.125 bits. So its is 65 or 66 bytes generally. 

To fix your issue, check that your coefficient is 66 bytes, and if its not, then append 00 to before it.

aseigler:
That makes sense.  Shouldn't be to hard to account for that.  Thanks @herrjemand!

herrjemand:
@aseigler Happy to help *)

PR #65, opened 2018-12-22
Removed net462 from test project

abergs:
Running ```dotnet test`` generates a bunch of errors if net462 is targeted.

### DIFF
--- a/fido2-net-lib.Test/fido2-net-lib.Test.csproj
+++ b/fido2-net-lib.Test/fido2-net-lib.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net462</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 


MeikTranel:
That's the error i told you about here: https://github.com/MeikTranel/MigrationSupport/issues/1#issuecomment-442956717

The issue is that .NET Standard 2.0 has APIs that .NET v4.6.2 doesn't have, but the .NET team still marked 4.6.2 as .NET Standard 2.0 compatible. One of the 43 APIs missing is the one you have here (https://github.com/dotnet/standard/blob/master/docs/comparisons/netstandard2.0_vs_net461/System.Security.Cryptography.md); which is why your test fails on .NET462. Because you produce .NET Standard packages this means that there can be consumers of your package using .NET 4.6.2, that will eventually hit this bug. That's why i added 4.6.2 in the first place.

MeikTranel:
I just checked. There's a nuget package that can help you avoid this scenario.

MeikTranel:
Well the dependency would only exist for the net46X depedency group, not for the main netstandard2.0 dependencygroup.

abergs:
I believe the failing test is due to #32

aseigler:
I will be fixing #32 soon. Just been considering the best way to do it.

aseigler:
This should be clean now.

abergs:
@MeikTranel Do you think we should avoid removing net462?

@aseigler It does passes in the CI now. Good work.

MeikTranel:
Sorry, christmas chores came first 😄 
We should at least try to make it sane. I was thinking we put a net461/net462 depedency group in there which has some msbuild targets erroring when you're trying to build it with net461 or net462.
Because the additional nuget dependency might not be such an easy task.

MeikTranel:
I just cooked something together locally. This appends a task to BeforeBuild that checks if the current targetframework is one of the offending frameworks and puts a big ol' error out.

We can put a help url and better text in there of course.

pcxp90kjez

abergs:
@MeikTranel Looks really good! Send a PR? :) And good idea to link to an issue at github (or a net462.md file) in the repo.

MeikTranel:
Will do but tomorrow tho. Already back into Christmas Dinner Madness V2 😂

PR #66, opened 2018-12-26
Added a guarding mechanism to prevent usage in .NET46X

MeikTranel:
Because of compatibility issues with .NET Standard 2.0 we have to make sure
there is a sane way for customers to learn that this package does not work with
.NET Framework 4.6.X Skus.


MeikTranel:
With this in, i'd see no issue with removing net46X from the targetframeworks property in the test projects. (That other PR)

abergs:
Good!

PR #69, opened 2019-01-08
Prefer enumeration

idaviddesmet:
Prefer the use of enumerations over strings since it avoids mistyping, plus you get IntelliSense.
The documentation also mentions the use of enumerations, for example:

https://w3c.github.io/webauthn/#attachment
https://w3c.github.io/webauthn/#attestation-convey


abergs:
First, Thank you for contributing.

Could you expand on why these enums are a better choice than the typed strings we already use?
https://github.com/abergs/fido2-net-lib/blob/master/fido2-net-lib/Objects/AuthenticatorAttachment.cs

Also, does using reflection cause any other issues?
Is there any perf impacts?

Also, could we remove the noise (changing BuildTargets) from this Pull so that we document every change correctly.

idaviddesmet:
Sure!
First of all, just to clarify, I've nothing against the TypedString class and neither I'm always in favor of enums. With that said, I think the purpose of doing the class way would be handy in scenarios where you want to do _something else_ which in this case is just about parsing the string and having a _token meaning_, where an enum IMHO is best suited at. This can be done as seen in the PR with built-in functionality using Enums (there's already a converter in the Json.NET library for handling those) and since the conversion is used merely for talking between the backend and the frontend, the value of having those in Enum would be reflected in other use cases.

For example, (excluding Enums can also be Flags) if there's a need to list the values it is easier to iterate over an Enum compared to creating an instance of each value string and adding those to an enumerable, it's also not a cheap operation when GC kicks in.

Speaking more about perf considerations, `String` comparisons are not as efficient as simply comparing the value of an `Enum` and if you take into equation storing those 'plainly' on the database there will be a [performance hit generally using varchar](https://stackoverflow.com/questions/2346920/sql-select-speed-int-vs-varchar) (due to size, collation...).

There are several articles that talks about the reasons of using either one, most are personal preferences but I think it all comes down to the use case, is it for giving a token a meaning or for something else?

idaviddesmet:
Could you further explain "remove the noise in changing the BuildTargets", please? Not sure about your meaning...

fido2-net-lib/Fido2NetLib.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="../lib/Chaos.NaCL/Chaos.NaCL.Include.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net472;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Unix'">netstandard2.0;net472;netcoreapp2.0</TargetFrameworks>
```
abergs: @idaviddesmet This change would be nice to remove from this PR

abergs:
@idaviddesmet I added an inline comment here: https://github.com/abergs/fido2-net-lib/pull/69/files#r246091298

fido2-net-lib/Fido2NetLib.csproj
idaviddesmet: Whoops! That must have slipped through...
I will remove it from this PR and create another one for it. This line is needed so it builds in macOS and Linux otherwise the compiler will argue about the missing .NET Framework 4.7.2 SDK in those platforms.

aseigler:
These three conformance tests now fail due to the PublicKeyCredentialType not having a default value of something other than "public-key".  These conformance tests send all kinds of nonsense including not sending a type value at all and are expecting the server to fail the operation.  Since PublicKey is the only option, if the client sends null type gets set to PublicKey and the test passes where the conformance tool expects to see a failure.

Server-ServerAuthenticatorAttestationResponse-Resp-1 Test server processing ServerAuthenticatorAttestationResponse structure

F-4 Send ServerAuthenticatorAttestationResponse that is missing "type" field and check that server returns an error

and 

Server-ServerPublicKeyCredentialGetOptionsResponse-Req-1 Test server generating ServerPublicKeyCredentialGetOptionsResponse

P-1 Get ServerPublicKeyCredentialGetOptionsResponse, and check that: (a) response MUST contain "status" field, and it MUST be of type DOMString…

and 

Server-ServerAuthenticatorAssertionResponse-Resp-1 Test server processing ServerAuthenticatorAssertionResponse structure

F-4 Send ServerAuthenticatorAssertionResponse that is missing "type" field and check that server returns an error

aseigler:
> is possible to integrate those tests?

In theory we could snag the raw json from a test run, save it as files and organize them into tests churning the json back in. Might be worth the effort to prevent future regressions.

idaviddesmet:
Made a PR #73 but I couldn't test it since I still don't have access to the conformance testing tool

> Your request was successful. Access may take up to one business day to be provisioned.

idaviddesmet:
> > is possible to integrate those tests?
> 
> In theory we could snag the raw json from a test run, save it as files and organize them into tests churning the json back in. Might be worth the effort to prevent future regressions.

I like the idea. [Project Wycheproof](https://github.com/google/wycheproof/tree/master/testvectors) uses such practice for the test vectors.

abergs:
> May I ask where did you got the tool and if is possible to integrate those tests?

Also @herrjemand is working on (or has it on the roadmap) making the conformance runnable from CLI

idaviddesmet:
> > May I ask where did you got the tool and if is possible to integrate those tests?
> 
> Also @herrjemand is working on (or has it on the roadmap) making the conformance runnable from CLI

But in order to integrate it in the CI pipeline will mean that such tool can be obtained from a public URL which currently you can't since you have to register unless I'm missing something? If added to the repo, wouldn't it violate their TOS?

aseigler:
Adding data from MDS would be a violation of the TOS for the access token (and yet I have seen it done) , and redistributing the conformance tool would probably be a no-no, but I don't believe the conformance tool output falls into the same category. It still might be frowned upon, I don't know.

PR #73, opened 2019-01-11
Properties of PublicKeyCredentialType enum are nullable

idaviddesmet:
Allow nulls in properties of PublicKeyCredentialType enum as mentioned [here](https://github.com/abergs/fido2-net-lib/pull/69#issuecomment-453638318).


abergs:
@aseigler can you run tests on this branch and see if this fixes the issue? (I'm unable at the moment)

aseigler:
I will try to get it tested tonight.

aseigler:
The two conformance tests related to this issue both pass acceptably with this fix applied.  The third one is actually not related to this update but a new requirement in the conformance test tool.

PR #74, opened 2019-01-27
Support Target Frameworks by OS

idaviddesmet:
A pending PR. Allows building in macOS and Linux otherwise, the compiler will argue about the missing .NET Framework 4.7.2 SDK in those platforms.


abergs:
Looks good. Merging.

PR #83, opened 2019-03-02
Convert metadata to async

aseigler:
Plus small fix to pass new Get Assertion F-5 test.

### DIFF
--- a/fido2-net-lib/MetadataService.cs
+++ b/fido2-net-lib/MetadataService.cs
@@ -741,19 +741,22 @@ namespace Fido2NetLib
         public MetadataTOCPayloadEntry GetEntry(Guid aaguid)
         {
             MetadataTOCPayloadEntry entry;
+            var retries = 0;
             if (true == ConformanceTesting())
             {
-                while (false == conformanceTOCReady)
+                while (false == conformanceTOCReady && retries < 60)
                 {
                     System.Threading.Thread.Sleep(1000);
+                    retries++;
                 }
                 ConformanceMetadata.payload.TryGetValue(aaguid, out entry);
             }
             else
             {
-                while (false == tOCReady)
+                while (false == tOCReady && retries < 20)
                 {
                     System.Threading.Thread.Sleep(1000);
+                    retries++;
                 }
                 mDSMetadata.payload.TryGetValue(aaguid, out entry);
             }
--- a/fido2-net-lib/MetadataService.cs
+++ b/fido2-net-lib/MetadataService.cs
@@ -310,6 +310,8 @@ namespace Fido2NetLib
     {
         MetadataTOCPayloadEntry GetEntry(Guid aaguid);
         bool ConformanceTesting();
+        bool IsInitialized();
+        Task Initialize();
     }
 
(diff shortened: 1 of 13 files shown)


fido2-net-lib/MetadataService.cs
```diff
@@ -316,58 +317,35 @@ public sealed class MDSMetadata : IMetadataService
...
-                // If the payload count is still zero and we have what looks like a good access token, load from MDS
-                if (0 == payload.Count && null != _accessToken && 0x30 == _accessToken.Length)
-                {
-                    GetTOCPayload(false);
-                }
-                // If the payload count is zero, we've failed to load metadata
-                if (0 == payload.Count) throw new Fido2VerificationException("Failed to load MDS metadata");
-            }
-            else AddConformanceTOC();
+            if (false == ConformanceTesting()) Task.Run(AddMDSTOC);
```
abergs: @aseigler I'm not completely sure but I feel some hesitation to create run a Task without support for awaiting and cancellation. I have an idea for this. What about removing the constructor for this class and let the consumer call a and await a Initialize method (if IsInitialized is false).

So: Let's add a two methods: `public async Task Initialize();` and a `public bool IsInitialized {get;}`

What do you think=

aseigler: Makes sense, definitely cleaner than this.

aseigler: This seems to work.  Caller can wait or timeout as they see fit, controllers currently should block until metadata is loaded.  Conformance test runs must block until all data is processed or tests will simply fail.

CodeTherapist: @aseigler I'm wondering from where do you have this inverted `if` style e.g.  `if (0 == payload.Count)`? I never saw that in a C# repo.

aseigler: It's an old habit to avoid accidentally assigning a value when you meant to compare. If that's not what the kids are doing these days I can try to reform.

CodeTherapist: @abergs I thought that comes from another programming language. For myself I'm used to the other style. The most important is, that everyone working on the repo is aware and agreed on a certain style :)

aseigler: I don't think style has ever come up, we've been too busy just trying to get this thing working and available.

abergs: As @aseigler says, we've been busy making it work. Also (while i'm accustomed to it) sometimes that inverted if-statements made the cert/crypto code more readable!

I added a ticket here: https://github.com/abergs/fido2-net-lib/issues/87

abergs:
LGTM.

I would change our controllers to be Async so that we can await the Initialize method if we want to.

@CodeTherapist So the reason why we don't initialize internally automatically is because we have sync API's and the init should be done async so to avoid complexity, we need to offload this.

Also running with MetaData is "kind of optional" and the consumer could use other MetaDataServices (cached on file for closed environments) etc.

abergs:
Go a head and merge @aseigler 👍

PR #90, opened 2019-04-15
Fixes enum parsing with dashed values

### LINKED ISSUE #89: ToEnum cannot parse "cross-platform"
The commit https://github.com/abergs/fido2-net-lib/commit/768512a1f864a9d9b555b6ed2ef92f62c4f4b076 introduced a bug which means that the option "cross-platform" cannot be parsed/used.

It throws the exception:
Value 'cross-platform' is not a valid enum name of 'Fido2NetLib.Objects.AuthenticatorAttachment' (ignoreCase=True). Valid values are Platform;CrossPlatform.

Could @CodeTherapist take a look?
Otherwise I suggest we revert that commit

idaviddesmet:
Fixes #89

### DIFF
--- a/fido2-net-lib/EnumExtensions.cs
+++ b/fido2-net-lib/EnumExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.Serialization;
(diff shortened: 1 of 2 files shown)


aseigler:
I like it. @abergs?

fido2-net-lib/EnumExtensions.cs
```diff
@@ -17,14 +18,24 @@ public static class EnumExtensions
...
-            {
-                return enumValue;
-            }
-            else 
+            // Try to parse it normally on the first try
+            if (Enum.TryParse<TEnum>(value, ignoreCase, out var result))
+                return result;
+
+            // Try with value from EnumMemberAttribute
+            foreach (var o in Enum.GetValues(typeof(TEnum)))
```
CodeTherapist: We could reuse the subsquent usage of `Enum.GetValues(typeof(TEnum))` on line 34 and 35.
But this is really minor, no reason to not merge it as is.

idaviddesmet: I also thought about it but wanted to avoid allocating the valid values, thinking it will be less common to fall into the exception and the allocation will be of no use when the loop returns the value.

CodeTherapist: Hmm, not sure I follow. But as soon you reach the `foreach` you allocating the valid values.
I didn't try that, but this should work:

```
            // Try with value from EnumMemberAttribute
            var values = Enum.GetValues(typeof(TEnum)).OfType<TEnum>().ToArray();
            foreach (var o in values)
            {
                if (ToEnumMemberValue(o).Equals(value, ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
                    return o;
            }

             throw new ArgumentException($"Value '{value}' is not a valid enum name of '{typeof(TEnum)}' ({nameof(ignoreCase)}={ignoreCase}). Valid values are: {string.Join(", ", values))}.");
        }
```

idaviddesmet: The returned valid values are converted using `ToEnumMemberValue` in order to spill out the strings like *cross-platform*.

CodeTherapist: Ah crap, yes sure. Then the solution would be:

     ```
           // Try with value from EnumMemberAttribute
            var values = Enum.GetValues(typeof(TEnum)).OfType<TEnum>().ToArray();
            foreach (var o in values)
            {
                if (ToEnumMemberValue(o).Equals(value, ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
                    return o;
            }

             throw new ArgumentException($"Value '{value}' is not a valid enum name of '{typeof(TEnum)}' ({nameof(ignoreCase)}={ignoreCase}). Valid values are: {string.Join(", ", values.Select(v => ToEnumMemberValue(n)))}.");
```
But as I already said, I'm fine with what you did.

idaviddesmet: The LINQ is internally doing a loop and allocating a new collection, but I have to agree that it looks more verbose and cleaner that way. 🙂 I will make the proper changes as suggested. Thanks!

Issue #79, opened 2019-02-12
[Proposal] Restructuring repository


CodeTherapist:
I propose to align the repository/solution to the following structure:

```
\
  Fido2.sln
  Directory.Build.props     (Common for all projects)
  \src
    Directory.Build.props   (Common for all src projects)
    \Fido2.csproj
    \Fido2.AspNetCore. ...
    \...
  \test
    Directory.Build.props   (Common for all test projects)
    \Fido2.Tests
    \Fido2.IntegrationTests
    \…
```

The reason why I propose it, is to help new developers that are used to [other dotnet projects](https://github.com/topics/dotnet) to get into the work (e.g. [aspnetcore](https://github.com/aspnet/AspNetCore)). A benefit from the build perspective is the [MSBuild built-in extensibility hook](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017#directorybuildprops-and-directorybuildtargets) that could be effectively used. Sure, this is also a matter of taste, rather than functionality. 

What do you think?

aseigler:
Makes sense, and we probably should have done this a long time ago.  I have been much more focused on passing conformance tests than external consumability.

CodeTherapist:
@aseigler If this is a "yes, do it" - then I would propose a PR accordingly :)

aseigler:
I will defer to @abergs as this is really outside of my area of expertise.

abergs:
LGTM.

However Demo should probably not be in src. In way Demo should be moved out from the library BUT it makes conformance testing a lot more tedious and harder, so i'm against it. But let's not place it in src.

CodeTherapist:
@abergs Could you please elaborate how would place the library into `src` lead to tedious conformance testing?

abergs:
Sorry. What I meant was that Demo should optimally not be in project/src (since it's not part of the library) and we should put it in project/demo/.

So we have:
```
Project/
  .sln
  demo/
  src/
  test/
```

CodeTherapist:
@abergs Got it! It‘s exactly what I proposed. Sorry, that I didn‘t outlined the demo project as well in the initial post. I will do a PR.

abergs:
If possible use git mv so that we don't nuke git blame on the files.

PR #95, opened 2019-07-02
Restructure repository per proposal in #79

aseigler:


aseigler:
@CodeTherapist does this resemble what you were looking to do?

CodeTherapist:
@aseigler Yes, is exactly what the my issue was about 👍  🎉

PR #98, opened 2019-07-12
use AppId instead of RpId to hash when AppId-Extensions-Flag is set

dgorbach:
https://www.w3.org/TR/webauthn/#sctn-appid-extension

> If true, the AppID was used and thus, when verifying an assertion, the Relying Party MUST expect the rpIdHash to be the hash of the AppID, not the RP ID.


aseigler:
Looks good, but there is one small issue.  Raw.Extensions is very commonly null, in which case Raw.Extensions.AppID crashes.

dgorbach:
of course. 
I'l submit another PR

PR #103, opened 2019-07-24
Moved Models and Public classes to Fido2NetLib.Models project

abergs:
I've made some major refactoring and moved a lot of the Objects namespace to a separate project, including the "public" objects such as Options, RawResponse classes. This is to help consumers to avoid getting unnecessary dependencies on the System.IdentityModel.tokens.Jwt package.

I noticed that package is a concern when implementing this is in a http://asp.net  non-core webapp. I wanted to use the OptionsClasses when sending JSON to clients, but the dependency on Tokens.JWT 5.* actually broke AspNet which requires 4.*

The problem could have been solved by me (as the consumer) by writing my own POCO-classes and mapping etc but that is a lot of boilerplate. I believe this approach will benefit everyone else on the OWIN/Asp.net stack

This PR is a W.I.P. Before merging we should consider renaming the Configuration class To Fido2Configuration since it is not a child class anymore.

All namespaces have been kept as-is. However Options.Create have been made public instead of internal and as wel as some property setters


abergs:
@MeikTranel Do you have any insights to share regarding splitting a library into two nuget packages but still have a easy development flow?

The way I see this being consumed is:

Flow:

* Developers install Fido2.Models into their ASP.Net project
* Developers install Fido2 into their Backend project.
* Developers populate data in models (from Fido2.Models) and calls function in Backend with said models.
Backend uses Fido2 to attest.

End result:

* Aspnet-project depends on Fido2.Models package
* Backend depends on Fido2 package
* Fido2 depends on Fido2.Models package

All this to avoid introducing the System.IdentityModel.tokens.Jwt version 5.* in aspnet projects.

While we work on the library, it would be preferable to have Fido2 depend on Fido2.Models **project** instead of **package** to make sure intellisense and f5 debugging is seamless. 

Any insights on this sort of setup?

abergs:
I solved it with a conditional in the csproj in https://github.com/abergs/fido2-net-lib/pull/103/commits/7d1258eaea33bbbca77c579933e99e9fbf5017d6.

I also added some ok-ish build scripts for publishing nugets.

Wondering if we should rename the new project to just Models before merging, thoughts @aseigler?

Models/Fido2NetLib.Models.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+  <PropertyGroup>
+    <PackageId>Fido2.Models</PackageId>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionSuffix>preview5</VersionSuffix>
```
MeikTranel: Are you sure you want these packages to have isolated versioning flows? Don't want you want FIDO2 and FIDO2.Models to have the same package version etc.?

Src/Fido2NetLib.csproj
```diff
@@ -23,17 +22,19 @@
     <PackageProjectUrl>https://github.com/abergs/fido2-net-lib</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/abergs/fido2-net-lib/blob/master/LICENSE.txt</PackageLicenseUrl>
   </PropertyGroup>
-
   <ItemGroup>
+    <PackageReference Include="Fido2.Models" Version="1.0.0-preview5" />
```
MeikTranel: If we were to release both packages as one deployment object but two artifacts this packagereference would not be required.

Models/Fido2NetLib.Models.csproj
abergs: The should have the same version. Currently building a .ps1 script to sync it up and release them.

Src/Fido2NetLib.csproj
abergs: I'm intrigued, could you explain more?

abergs:
I believe my assumption to move dependency to different projects would allow side-by-side assemblys is wrong and there is by defualt no side by side. This is what I thought would allow multiple versions of System.identitymodel.token.jwt to co-exist.

Project A (asp.net full framework)
- dependent on System.IdentityModel.Tokens.Jwt 4.*
- dependent on Fido2.Models
- dependent on Project B

Project B
- dependent on Fido2

Fido2
- dependent on Fido2.Models
- dependent on System.IdentityModel.Tokens.Jwt 5.*

Asp.net full framework has a dependency on Jwt 4.*, so if we want to offer Fido2 to those (We should) we need to come up with a solution.

An idea I have is to condition our dependency on Identitymodel to either 4* or 5* depending on targeting framework. This comes with the price that we need to re-implement parts of our code that is using JWT because of the breaking changes in JWT, dependent on targeting framework.

Other ways (maybe?) could be to guide consumers to step 2 or 3 here: https://michaelscodingspot.com/how-to-resolve-net-reference-and-nuget-package-version-conflicts/

abergs:
Or - crazy idea - we source the parts for validating a token as we've done with other projects. The code is MIT.

Parsing: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b0867057c1b98ead823eed37869b5656ed2ba695/src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs

ValidateToken:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e69d0c77909334ddba19ac15a6d993adcc77e458/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs

Do you think this would be doable @aseigler ?

Src/Fido2NetLib.csproj
MeikTranel: With csprojs doing "dotnet pack" on a csproj will automatically pack all projects referenced as well while adding nuspec dependencies to them simultaneously (unless specifically disabled of course).
With a shared Version variable this means that dotnet pack on FIDO2 will pack both Fido2.nupkg and Fido2.Models.nupkg with the versions and dependencies automagically set.

MeikTranel:
> Or - crazy idea - we source the parts for validating a token as we've done with other projects. The code is MIT.
> 
> Parsing: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b0867057c1b98ead823eed37869b5656ed2ba695/src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
> 
> ValidateToken:
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e69d0c77909334ddba19ac15a6d993adcc77e458/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
> 
> Do you think this would be doable @aseigler ?

Is JWT-Communication a center pillar of your current implementation? Or is just a flavor of usage - is another alternative usage possible?
If its the latter i would just do "Fido2" and "Fido2.JWT" wherein JWT has all the types required to configure Fido for usage with JWT.

abergs:
It's unfortunately used to verify AndroidSafetyNet and the whole MDS chain so moving it to another package would not be recommended.

I made a quick attempt to depend target NET471 and then depend on the JWT 4.* package, but then it seems like we loose the ECDsaSecurityKey class.

I'm running out of ideas. If possible i think sourcing is the most suitable way to support asp.net full framework clients - although I'm not sure how feasible that is.

Src/Fido2NetLib.csproj
abergs: Sounds great!

MeikTranel:
One thing is sure tho. Sourcing types from microsoft packages is bound for disaster if you're making it an exchange type... you can not for all that is holy expose this type to the public interface of your package.

abergs:
@MeikTranel It would only be kept internal to parse the incoming JWT bytes. Still feels like a last resort.

Tratcher:
> but the dependency on Tokens.JWT 5.* actually broke AspNet which requires 4.*

Microsoft.OWIN 4.0 also moved to Tokens.JWT 5.x. Are you still targeting Microsoft.OWIN 3.x? 3.x is out of support.
https://www.nuget.org/packages/Microsoft.Owin.Security.ActiveDirectory/

aseigler:
I can look into moving the JWT stuff to a different package if it would help, but I don't know what would be most compatible with what you're talking about doing.

abergs:
What @Tratcher is saying is interesting though. The App I'm testing with has Owin-packages 4 installed... but still breaks down.

abergs:
@Tratcher It seems like that SaaS app was partly running on owin4.0 with bindingRedirect hell as well as a dependency on IdentityServer3 packages that required Tokens.JWT 4.*. Once that was cleared up it worked fine.

While it's a bit of a shame we can't support Owin < 4.0 out of the box there seems to be OK workarounds.

MeikTranel:
I'm about to through you a PR on this branch.

PR #105, opened 2019-07-26
Wrong naming of attestationObject and clientDataJSON in usernameless.…

Wesseldr:
…register.js

During implementation and using this lib, i found these two anomalies vs the specs.
Lookup up the names on the specs: https://www.w3.org/TR/webauthn/#iface-authenticatorattestationresponse
AttestationObject -> attestationObject
and
clientDataJson -> clientDataJSON

PS I did not check on the server side (I'm not a .net dev),  if the server is handled correctly or if that needs an update too. My guess is it needs. Could some .NET wiz please check this and update the code if things fail at the back end? Many thanks in advance.


Wesseldr:
PS I did not check on the server side if that is handled correctly or if that needs an update too.

PR #107, opened 2019-07-29
Out of spec typo on clientDataJson -> clientDataJSON

Wesseldr:
Specs on https://www.w3.org/TR/webauthn/ 
use the clientDataJSON not the clientDataJson
See change in file. This is in the demo usernameless.register.js
I know it's just details but it's an API the W3C docs are pretty strict in ;-)


Wesseldr:
This is in the demo usernameless.register.js

abergs:
Thanks, Fixed in ee52d89a57b6d96

PR #108, opened 2019-07-29
Add field userHandle in verifyAssertionWithServer

Wesseldr:
For the usernamaless result on the client side, the **userHandle** was missing in the reply. 
This can be _nullable_, but not missing since it is even backwards compatible according to: https://www.w3.org/TR/webauthn/#conforming-authenticators-u2f

however: FIDO2 BtB (by the book ;-)
Quote from: 
https://www.w3.org/TR/webauthn/#discover-from-external-source 
step: "->If any authenticator indicates success."

**userHandleResult**
If the authenticator returned a _user handle_, set the value of **userHandleResult** to be the bytes of the returned user handle. Otherwise, set the value of **userHandleResult** to **null**.

I'm no .NET guy, so haven't checked what impact this field has on the backend when it suddenly appears. I presume 'nothing' but not sure. If someone else could please test that out for me. Many thanks!


abergs:
I do think that part of the spec is for UserAgents, but never the less this field should be populated.

PR #111, opened 2019-07-29
Shuffled around props to reduce complexity and increase flexibility

MeikTranel:
Also fixed the documentation URL in the net46x warning
Suppressed warnings in sourced libs
Globalized supported targetframeworks as well as version variables etc.

After merging we have to adapt build scripts again cause we don't need to sync version etc.

Everything can be done by running `dotnet build/test/pack` on the solution file / in the root directory where the solution lies.
Packages will automatically have synced up dependencies. This should help you peeps concentrate on the important stuff. If you've come to a conclusion regarding how the packages should be structured i can help finalize all this again.

Short suggestion on my end:

- Do not get anxious about adding dependencies -  chances are consumers of Fido2 will have to use packages anyways.
- It sounds like you would want client and core stuff in the Fido2 package and another package for adding servicing capabilities in ASP.NET Core. I propose you actually call this package Fido2.AspNet or something like that.


abergs:
Looks good! Thanks a lot @MeikTranel 

Regarding aspnet servicing, I think you're correct. That package could contribute middlewares to make life easier for any ASP.NET Core consumers.

I do think that splitting models and core is still in our interest, regardless of ASP.NET Core package. Any projects that are only interested in the Models (e.g api frontends) wouldn't be sucking in unnecessary dependencies. 

Will merge tomorrow.

PR #118, opened 2019-08-31
corrected error variable

Celdus:
I just changed the catched error variable from "e" to "err", so it gets used in the line below to show the error message.


aseigler:
Thanks!

abergs:
Thanks!

PR #119, opened 2019-09-01
Problem with non English OS

fa18swiss:
The comparison for the X509BasicConstraintsExtension and the Extended Key Usage 2.23.133.8.3 doesn't work when the OS is not in English (usage of FriendlyName). I have fixed and validated with the units tests.

I have also added missing using blocks


abergs:
LGTF, but I want @aseigler to verify the certificates part

PR #125, opened 2019-09-22
MDS refactoring

mackie1001:
As discussed briefly a little while back.

Refactored metadata related to services with the following goals:
- Separate the concerns of the various sources of data: FIDO2 MDS, static, filesystem and conformance testing
- Provide caching support leveraging standard infrastructure in ASP.Net Core
- Pleasant developer experience
- Intuitive DI behavior, reduced reliance on static singletons
- No breaking changes within the existing core library (i.e. keep IMetadataService the same)

The result is:
- IServiceCollection fluent extensions to guide registration (see Fido2NetLibBuilderExtensions)
- Simple and caching (via IDistributedCache) implementations of IMetadataService that can accept a collection of priority-ordered metadata repositories. Cache aligns expiry with MDS TOC expiry.
- All additional dependencies are in a new Fido2.AspNet project that should be easy to package up in nuget


abergs:
I like this. However, build is failing due to:

error CS0246: The type or namespace name 'ServiceCollection' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\Test\Test.csproj]

Would you mind fixing those last errors so that it works as intended and I will do a full checkour review for it?

I appreciate your work, thank you.

mackie1001:
Sure, no problem. Will sort this out in the next day or two.

mackie1001:
FYI,  I tested that the conformance testing client could initialize properly but haven't attempted to run the conformance testing tool against the demo app.

aseigler:
Part of the conformance testing requires downloading some metadata statements from the conformance application and pre-populating the conformance statement cache with those files out of the file system, to demonstrate that the server is capable of reading metadata statements that are not included in MDS.  How can we do that with IDistributedCache?

aseigler:
[metadata.zip](https://github.com/abergs/fido2-net-lib/files/3642812/metadata.zip)

Attaching actual conformance statements that need to be pre-loaded into the cache.

mackie1001:
I think I want to add some logic to deal with the possibility that the TOC "NextUpdate" value could be in the past if we assume the time is "00:00:00" when parsing the date and the cache service will throw an exception if the absolute expiry time is in the past. Watch this space.

aseigler:
@mackie1001, I made a few changes to get the conformance testing to work again and sent a PR to your fork.

aseigler:
LGTM!

Issue #87, opened 2019-03-11
Add Coding Style to Contributing.md and editorconfig


abergs:
Took this from the ASP.NET Core repo.

I think we could apply the same rules to our code. Also a .editorconfig would be nice.

## Coding guidelines

:grey_exclamation: The *content* of the code that we write.

### Coding style guidelines – general

The most general guideline is that we use all the VS default settings in terms of code formatting, except that we put `System` namespaces before other namespaces (this used to be the default in VS, but it changed in a more recent version of VS).

1. Use four spaces of indentation (no tabs)
2. Use `_camelCase` for private fields
3. Avoid `this.` unless absolutely necessary
4. Always specify member visibility, even if it's the default (i.e. `private string _foo;` not `string _foo;`)
5. Open-braces (`{`) go on a new line
6. Use any language features available to you (expression-bodied members, throw expressions, tuples, etc.) as long as they make for readable, manageable code.
   * This is pretty bad: `public (int, string) GetData(string filter) => (Data.Status, Data.GetWithFilter(filter ?? throw new ArgumentNullException(nameof(filter))));`

### Usage of the var keyword

The `var` keyword is to be used as much as the compiler will allow. For example, these are correct:

```c#
var fruit = "Lychee";
var fruits = new List<Fruit>();
var flavor = fruit.GetFlavor();
string fruit = null; // can't use "var" because the type isn't known (though you could do (string)null, don't!)
const string expectedName = "name"; // can't use "var" with const
```

The following are incorrect:

```c#
string fruit = "Lychee";
List<Fruit> fruits = new List<Fruit>();
FruitFlavor flavor = fruit.GetFlavor();
```

### Use C# type keywords in favor of .NET type names

When using a type that has a C# keyword the keyword is used in favor of the .NET type name. For example, these are correct:

```c#
public string TrimString(string s) {
    return string.IsNullOrEmpty(s)
        ? null
        : s.Trim();
}

var intTypeName = nameof(Int32); // can't use C# type keywords with nameof
```

The following are incorrect:

```c#
public String TrimString(String s) {
    return String.IsNullOrEmpty(s)
        ? null
        : s.Trim();
}
```

### Cross-platform coding

Our frameworks should work on CoreCLR, which supports multiple operating systems. Don't assume we only run (and develop) on Windows. Code should be sensitive to the differences between OS's. Here are some specifics to consider.

#### Line breaks
[...]

PR #127, opened 2019-09-29
Code cleanup

Shtong:
Here is a first working draft to address #87. It's getting pretty big, since in addition to editing `.editorconfig` to the rules we talked about, I also started to apply the rules to the code.

A few things to note however:
- I have done my best to not touch anything that would have broken compatibility for the library users. So for example, public async methods were not renamed to have an Async suffix.
- I have sometimes made changes to follow the [official capitalization rules for acronyms](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN#capitalization-rules-for-identifiers): for example, `conformanceTOCReady` was renamed `_conformanceTocReady` (with lowercase 'oc')
- In some places I took the liberty to replace `ArgumentException` (or an equivalent) with `InvalidOperationException`, which is the correct exception that should be used when the state of an object (and not an argument) is causing a problem.

I'm not sure about the end result to be honest. Because I wanted to not break the public interface, it looks half-baked. Also because of the very large amount of small changes, the PR looks very hard to review. I feel it should either be less ambitious, or go all the way and break the existing API in the process, maybe producing a new major version. Any thoughts ?


aseigler:
This is actually not too bad to review, as all the code is extremely familiar to me.

As far as the capitalization rules go, in many cases I used the abbreviations directly from the WebAuthn spec, various FIDO specs, or in the case of TPM attestation, TCG specs.  In the specific example you gave, in the FIDO specs the metadata [table of contents](https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-metadata-service-v2.0-rd-20180702.html#metadata-toc) is always referred to as a TOC, so I tried my best to match the case from the spec exactly, so that someone trying to map the code in the library to the spec would find it very easy to follow.  If that's not proper, that's fine, we can name things however the cool kids are doing it these days.  So long as it works and makes sense, I am ok with that.

Src/Fido2/AttestationFormat/AndroidKey.cs
```diff
@@ -21,25 +23,34 @@ public static byte[] AttestationExtensionBytes(X509ExtensionCollection exts)
...
+                throw new Fido2VerificationException("attExtBytes signature not a valid DER sequence");
+
             // next is length of all the items in the sequence
             var dataLen = AuthDataHelper.GetSizedByteArray(attExtBytes, ref offset, 1);
-            if (null == dataLen) throw new Fido2VerificationException("attExtBytes signature has invalid length");
+            if (null == dataLen)
+                throw new Fido2VerificationException("attExtBytes signature has invalid length");
+
             // if data is more than 127 bytes, the length is encoded in long form
+            // TODO : Why is longLen never used ?
```
aseigler: Because I forgot to bring over the sanity check added on that field I added in #100.

aseigler:
LGTM but we introduced some conflicts by pulling #125. Should not be too bad to resolve.

Shtong:
Alright. I started merging; although I want to make a pass on the many new files so they use the same rules as the other ones I already refactored before we finish that PR.

aseigler:
Thank you!

Issue #126, opened 2019-09-26
Use an Azure Pipeline configuration file


Shtong:
**Suggestion**: Use YAML Pipelines to build and deploy the project, instead of the classic web interface configuration

**Reasoning**: Azure Pipelines now provides (and recommends) a new way to configure the build and release process. It uses a YAML file stored in the built project's repo, instead of defining it using the DevOps web interface (now called "classic" mode).

Using this mechanism has several advantages : 
- It allows contributors to update the build process as part of their pull request (for example it would have been useful in #124 to update the target SDK version)
- Branches can use a build configuration that is specific to them, without breaking the build for the other branches
- The full build configuration is now in a file in the GIT repo, making it easier to access

abergs:
I agree with the reasoning. I believe the configuration can be extracted from the interface.

Shtong:
It is possible. The result usually needs a little bit of cleanup but it's pretty simple overall!

Shtong:
I've tried to have a look at this but I don't have the permissions to look at and export the pipeline's YAML files. So someone with enough permissions will have to do it :)

abergs:
Exported YAML:

#


# Fido2 Demo Build

```
pool:
  name: Hosted VS2017
#Your build pipeline references the ‘BuildConfiguration’ variable, which you’ve selected to be settable at queue time. Create or edit the build pipeline for this YAML file, define the variable on the Variables tab, and then select the option to make it settable at queue time. See https://go.microsoft.com/fwlink/?linkid=865971

steps:
- task: NuGetToolInstaller@0
  displayName: 'Use NuGet 4.3.0'

- task: DotNetCoreInstaller@0
  displayName: 'Use .NET Core sdk 2.2.101'
  inputs:
    version: 2.2.101

- task: DotNetCoreCLI@2
  displayName: 'dotnet restore'
  inputs:
    command: restore

- task: DotNetCoreCLI@2
  displayName: 'dotnet build entire sln'
  inputs:
    command: publish
    arguments: '--configuration $(BuildConfiguration) --output $(build.ArtifactStagingDirectory) --framework netcoreapp2.2'

- task: PublishBuildArtifacts@1
  displayName: 'Publish Artifact: demodrop'
  inputs:
    ArtifactName: demodrop
```



# Fido2-CI
```
pool:
  name: Hosted VS2017
steps:
- task: NuGetToolInstaller@0
  displayName: 'Use NuGet 4.3.0'

[...]

PR #124, opened 2019-09-22
Preparations for .NET Core 3

Shtong:
This PR contains changes to prepare the library to work with .NET Core 3.
- The website has been updated to use ASP.NET Core 3.0 instead of ASP.NET Core 2.2
- The FIDO2 library now also targets `netcoreapp3`

I haven't changed the library code, and everything seems to be in order.

Note: I plan to update this branch again once the final versions of the 3.0 dependencies are released. This PR can be reviewed, as there won't be much changes needed to switch to the final dependencies, but please wait until I've done the update before you merge :)

PR #143, opened 2020-02-18
Upgrading project to remove EOL frameworks

ewassef:
Updating the application to .net core 3.1 based on the guidelines defined [here](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)

This updates the libraries to .netstandard 2.1 but maintains .netstandard 2.0 for those using this library with .netcoreapp 2.1

this will also rev the version from 1.0.1 to 1.0.2 to indicate a patch increment (even though the framework was updated) as the package is still backwards compatible and no new features have been added

### DIFF
--- a/Demo/Demo.csproj
+++ b/Demo/Demo.csproj
@@ -7,9 +7,8 @@
     <ProjectReference Include="..\Src\Fido2.AspNet\Fido2.AspNet.csproj" />
     <ProjectReference Include="..\Src\Fido2.Models\Fido2.Models.csproj" />
     <ProjectReference Include="..\Src\Fido2\Fido2.csproj" />
-
-    <PackageReference Include="Microsoft.AspNetCore.All" />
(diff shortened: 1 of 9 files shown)


ewassef:
@abergs  The pipeline step that's breaking is explicitly trying to build in .netcore framework 2.2 (this will always break since this PR is removing that). 
Is there a particular reason the framework was specified?

abergs:
@ewassef The pipeline configuration needs a version to, however as discussed in #124 and #126 that is due to change soon.

ewassef:
That makes a ton of sense. Do you know when something like this will happen? We are wanting to start using this in .net 3.1 as soon as we can

ewassef:
What would you recommend in regards to the timeline? Should we fork and attempt a merge in the future?

abergs:
Done in #145.

Are there any other changes you want to make?

ewassef:
Nope! thank you for doing this. I will try to pull the latest and incorporate them in. If I run into anything else, ill submit a fix!

PR #152, opened 2020-04-15
Bug fix: require_resident_key

Shane32:
The `require_resident_key` property does not work properly unless it is a string.  This PR fixes that bug.  Tested with Chrome on Windows 10.


Shane32:
It would be nice if someone else could test this patch within other environments.  In my environment, the effect of the change is that the pin code is not required when using a physical security token on the 'passwordless' page.  You can compare the demo site to https://webauthn.io/ where they are passing the `require_resident_key` as a string value, and it does not ask for a pin for passwordless authentication.

Issue #149, opened 2020-03-24
Create and automate Release notes


abergs:
Since we now have a stable version of Fido2 published we need to maintain release notes to help developers upgrade.

* Automate as much as possible (releases are currently done with `scripts/run-all.ps1`
* Add information to nupkg release notes
* Add information to GitHub Releases
* Git tag support?

abergs:
I believe this Task could be helpful, will explore: https://devblogs.microsoft.com/devops/automating-releases-in-github-through-azure-pipelines/

abergs:
Alright, successfully added Release Drafter (https://github.com/marketplace/actions/release-drafter) which generates a draft release for every merge on the repo, result looks like this. It also have good support for categories such as bug fixes/features depending on labels on the PR.

So current flow is:

* Manually bump version numbers (using `Scripts/setVersion.ps1`).
* Create PR with perhaps discussion regarding Major.Minor.Patch, sharing the Draft Release notes.
* Merge PR. Publish Release notes.
* Checkout tag, run `Scripts/run-all.ps1` to build, setVersion and push to nuget.

Could be improved to:
* Semi-Manually (script) bump version number, Create PR where Release Notes and version can be discussed and changed.
* Merge PR, automatically publishes release notes, builds and pushes to nuget.

PR #160, opened 2020-05-09
Adding a changelog, and 2 links to the readme

damienbod:
Firstly thanks for the great project. I think it would help if this project had a changelog, and made it easier to find when what was released and where you can find the nugets, releases. This PR should help the visibility a small bit. 

I think also it would be great if you added release tags for the public releases. (Version 1.0.1 and Version 1.1.0) Then you could create a github release in your repo using the release tab from this repo. Since you don't release so often, you can do this manual, or course this could be automated using your pipeline. 

Just some ideas, 

Thanks for the project.

Greetings Damien

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 [Readme](https://github.com/abergs/fido2-net-lib/blob/master/README.md) 
 
-2020-04-24 1.1.0
+2020-03-24 1.1.0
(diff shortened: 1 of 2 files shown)


abergs:
Since the overhaul of release notes in #149 I'm going to delete the changelong.md and incorporate it in release notes since they will be kept up to date moving forward

Issue #153, opened 2020-04-17
Upgrade to System.IdentityModel.Tokens.Jwt 6.5.0 has broken the Fido2MetadataServiceRepository


mackie1001:
It looks like the implementation of `JArray` used by `JwtSecurityToken` has been switched to an internal non-Newtonsoft one as of the latest version and this has broken `DeserializeAndValidateToc` which casts JWT properties to that type. 

This crept in because I updated another package that brought in 6.5.0. The version currently referenced in 5.6.0.

Want me to do a PR for this?

abergs:
Thanks for the report! When does this problem manifest itself, do you have a repro?

mackie1001:
Basically if I reference `fido2-net-lib` but upgrade all packages (I depend on System.IdentityModel.Tokens.Jwt elsewhere) to latest via the package manager then it breaks as it binds against the newer version at runtime. This came about as I'm updating my current solution (which was using a version pre-my PR) to .Net Core 3.1 and updating all other dependencies to match.

I've worked around it for now by explicitly referencing 5.6.0 in my solution.

mackie1001:
PR #156 created for this.

aseigler:
This same problem also impacts SafetyNet attestation type verification as it uses JArray as well.

aseigler:
This is fixed, verified.  Thanks!

PR #165, opened 2020-05-21
Fix SafetyNet attestation verification, add unit tests

aseigler:
- Related to #153, SafetyNet attestation verification broke when System.IdentityModel.Tokens.Jwt was updated.  Fixed in similar fashion.

- Refactor SafetyNet attestation verification

- Add SafetyNet attestation verification unit tests

- All 172 conformance tests pass with conformance tool version 1.3.4.


aseigler:
image

Issue #167, opened 2020-05-27
Request for new nuget release with recent PRs


mackie1001:
I'm in the middle of a 3.1 upgrade project and would like to upgrade all packages to latest. Is it possible to get a new nuget release some time soon please?

abergs:
Yes @mackie1001, we should be able to release a new version. I could certainly release a preview package if that would help you get things moving?

I'm facing a problem with releasing new versions that I want to tackle, and I welcome all advise and help.

The problem is simply popularity. The project has matured, people are using the library (which is great!) which now means the releases needs to be properly managed. SemVer, Release Notes etc.

I don't have the resources to put together proper Release Notes, so I would like to automate and crowd source as much as possible. This would also help future proof the process, so that I'm not the only one at the wheel.

If anyone are able to lend their hand, please look into #149

abergs:
@mackie1001 See: https://github.com/abergs/fido2-net-lib/pull/175

mackie1001:
Nice work, thanks.

PR #168, opened 2020-05-27
Upgrade PeterO.Cbor package to v4.1.2

aseigler:
...and switch to using it as a PackageReference.  This makes more sense to me than continuing the way this has been configured, but I'm open to comments and suggestions.


abergs:
The reason why it's not a PackageReference: https://github.com/abergs/fido2-net-lib/issues/7

Isn't the concern still valid?

aseigler:
The concern hasn't changed, I am just not completely convinced the concern was ever valid to begin with, and wanted to bring it back up for discussion since many more eyes are on this now. The project has other package references, including those where upgrades have broken stuff, like #153. Seems like using a package reference would make it easier to keep dependencies up to date, see #167. It just feels odd to have a robust package management system and then completely circumvent it in the manner we have, that's why I wanted to bring it back up.

abergs:
I hear you, maybe it's over-the-top.

aseigler:
Merging based on https://twitter.com/blowdart/status/1266020693558169600.

Issue #185, opened 2020-08-27
Unit test failure when running tests locally


kennep:
When running the unit tests locally, 7 unit tests under "Packed" fail. However, I see that the same tests run successfully on Azure DevOps. So _something_ is different here.

What I've tried:
* `dotnet test` on Linux
* Running the tests in Rider on Linux
* Running the tests in Visual Studio 2019 on Windows

In all these three cases, the same 7 tests fail.

Environment:
```
$ dotnet --list-runtimes
Microsoft.AspNetCore.App 3.1.7 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
Microsoft.NETCore.App 3.1.7 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
$ dotnet --list-sdks
3.1.401 [/usr/share/dotnet/sdk]
```

Test log:
```
$ dotnet test
Test run for /home/kenneth/Projects/Vipps/fido2-net-lib/Test/bin/Debug/netcoreapp3.1/Test.dll(.NETCoreApp,Version=v3.1)
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
[xUnit.net 00:00:09.42]     Test.Attestation.Packed.TestFullAttCertNotV3 [FAIL]
[xUnit.net 00:00:09.43]     Test.Attestation.Packed.TestFullInvalidAlg [FAIL]
[xUnit.net 00:00:09.43]     Test.Attestation.Packed.TestFullAttCertSubject [FAIL]
  X Test.Attestation.Packed.TestFullAttCertNotV3 [8ms]
  Error Message:
   Assert.Equal() Failure
                 ↓ (pos 7)
Expected: Packed x5c attestation certificate not V3
Actual:   Packed signing certificate expired or not yet va···
                 ↑ (pos 7)
  Stack Trace:
     at Test.Attestation.Packed.TestFullAttCertNotV3() in /home/kenneth/Projects/Vipps/fido2-net-lib/Test/Attestation/Packed.cs:line 893
  X Test.Attestation.Packed.TestFullInvalidAlg [7ms]
  Error Message:
   System.AggregateException : One or more errors occurred. (Assert.Throws() Failure
Expected: typeof(System.InvalidOperationException)
[...]

PR #186, opened 2020-08-27
Compare certificate validity period correctly when the machine is using a time zone other than UTC

kennep:
When comparing two DateTimes, the `Kind` attribute is ignored, so comparing an UTC DateTime and a local DateTime do not give the expected result.

While the validity period is encoded as UTC in the certificate itself, it is converted to local time before it is returned in the X509Certificate2.NotBefore / X509Certificate.NotAfter attributes. This means that if we want to check whether the certificate is currently valid, we should compare it with the current _local_ time.

Fixes: #185


abergs:
Really sneaky bug, good catch 🐛

PR #188, opened 2020-08-29
MDS error handling and cache-friendlieness refactoring

mackie1001:
Improved error handling and logging for MDS errors along with a refactoring of how the TOC JWT alg is passed around to better serve the cached use-case.

The the TOC alg now gets stored in the MetadataTOCPayload class which is now passed into GetMetadataStatement. This means that it's possible to use a cached MetadataTOCPayload not created by the current IMetadataRepository instance.. 

Together these changes mean that an incident like we had with the Yubico version number won't cause the app startup to fail and will be recoverable even if the TOC is already cached.

This change will require anyone using the DistributedCacheMetadataService to clear their cache when upgrading to 2.0.

### DIFF
--- a/Demo/Startup.cs
+++ b/Demo/Startup.cs
@@ -53,11 +53,12 @@ namespace Fido2Demo
             .AddCachedMetadataService(config =>
             {
                 //They'll be used in a "first match wins" way in the order registered
-                config.AddStaticMetadataRepository();
+                
                 if (!string.IsNullOrWhiteSpace(Configuration["fido2:MDSAccessKey"]))
                 {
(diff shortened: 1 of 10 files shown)


aseigler:
@mackie1001, as of now we no longer need any of the static metadata as all of the authenticators we have in there are in the production MDS.  Do you want to remove that as part of this as well?

mackie1001:
@aseigler Yes, that makes sense. I recently ripped out my fake repo in my application too, the only downside being the icons in the MDS are not really suitable for presentation in our UI :(

aseigler:
Conformance metadata has moved

From: https://fidoalliance.co.nz/mds/
To: https://mds.certinfra.fidoalliance.org

PR #201, opened 2020-11-25
Fix typo in exception message - attestion -> attestation

ondrokrc:


ondrokrc:
Closed - will create again without the unnecessary commit.

Issue #146, opened 2020-03-24
Migrate from Newtonsoft to System.Text.Json


maxkuzmin:
Since in asp.net core 3 internal support of Newtonsoft is removed, now we need to use `.AddNewtonsoftJson();` to make this lib working. It seems quite redundant. Is it possible to migrate this library from Newtonsoft to System.Text.Json?

abergs:
@maxkuzmin It needs to be investigated, it's not as easy as it sounds.

System.Text.Json is optimized for speed and therefore cut features to obtain that speed.

Because of how WebAuthn works we need to pass base64 strings and rename models, e.g.

```csharp
 [JsonProperty("userHandle")]
 [JsonConverter(typeof(Base64UrlConverter), Required.AllowNull)]
 public byte[] UserHandle { get; set; }
```

abergs:
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to

PR #204, opened 2020-12-15
Migration to System.Text.Json

luanmm:
Hello, people.

This is **an initial test** on the update from Newtonsoft.Json to System.Text.Json. I just changed the Fido2.Models project and made it build - as it is the simplest project to start.

There is some tricky points, as stated in issue #146, because the project uses some attributes from Newtonsoft.Json that are not present in the System.Text.Json.

I'm not sure if they are really needed, as most of it could be replaced by some global options like IgnoreNullValues or something similar (and it does not affect functional aspects of the library anyhow, where I can see at least).

I've made it to discuss with you what do you think about moving forward with this PR. Maybe, if you find it viable, I could help on it (not sure how big this change will be, so it depends if I could make some time here to work on this).

Anyway, thanks for all the work that you've spent in this library. It is something really great for the authentication future of the web and apps.


abergs:
Hello, thank you for your pull request!

I'm not updated with the current situation of System.Text.Json.
How will this affect consumers not running on .net 5 or asp.net core 3.1?
We want this library to be backward compatible, and I suspect system.text.json might not be?

Is this something we might want to have multitargeting for?

luanmm:
Thanks for the reply, @abergs!

As I've been digging around, it should work in projects that aren't targeting .NET Core aswell (as pointed out in some threads like this one: https://github.com/dotnet/runtime/issues/31462). But before doing some work on this, some tests regarding compatibility would be nice too.

olmobrutall:
I would be very interested in this feature to get merged. Is there anyway to make it work out of the box in a .Net 5 project using System.Text.JSON otherwise?

aseigler:
Right now, full migration is blocked by two things:

1. https://github.com/dotnet/runtime/issues/31081, specifically for PublicKeyCredentialType and AuthenticatorAttachment.  Possible workarounds?
2. JObject usage, mainly in AndroidSafetyNet areas but also in some of the metadata processing areas, needs to be migrated to JsonDocument and that doesn't look super trivial.

PR #208, opened 2021-03-03
change origin verify to use fully qualified origin

dgorbach:
A Huawei P40 Lite sets the origin on register or sign in to https://www.passwordless.dev/custom#heroFoot and the verification fails.

With this PR,  the fully qualified origins are compared. see also https://github.com/duo-labs/webauthn/pull/60


aseigler:
LGTM, thanks!

Issue #226, opened 2021-04-23
Windows Hello only working with attestation type 'none' since version 2.0


m-gug:
Hello!
I updated to version 2.0 an noticed, that Windows Hello is no longer working with attestation type 'direct' or 'indirect'.
MakeNewCredentialAsync throws an error "Attestation with full attestation from authenticator that does not support full attestation".
Looks like the AttestationTypes for '08987058-CADC-4B81-B6E1-30DE50DCBE96' in the StaticMetadataRepo was changed from (ushort)MetadataAttestationType.ATTESTATION_BASIC_FULL to 1588 [see commit 1c80ae0c951cec008324a523500e74d311f9b05a](https://github.com/abergs/fido2-net-lib/commit/1c80ae0c951cec008324a523500e74d311f9b05a#diff-de3ccd7abcceb7fe8312f269b73f75853baf4c5de3b8b66235f5418c4af1d31b)

PR #227, opened 2021-04-30
Changed the AttstationType for Windows Hello Authenticator back to Full

m-gug:
With the current AttestationType Value of 1588 it's not possible to register a new Windows Hello Authenticator when requesting 'direct' oder 'indirect' because the certificate is not self-signed. The AuthenticatorAttestationResponse fails with an Exception 'Attestation with full attestation from authenticator that does not support full attestation'. #226 
Using ATTESTATION_BASIC_FULL instead of 1588 fixes the problem.

The value was changed in a previous [commit](https://github.com/abergs/fido2-net-lib/commit/1c80ae0c951cec008324a523500e74d311f9b05a#diff-de3ccd7abcceb7fe8312f269b73f75853baf4c5de3b8b66235f5418c4af1d31b), guess there was a reason for that change, maybe you can check if my change breaks something.

### DIFF
--- a/Src/Fido2/Metadata/StaticMetadataRepository.cs
+++ b/Src/Fido2/Metadata/StaticMetadataRepository.cs
@@ -214,7 +214,7 @@ namespace Fido2NetLib
                 {
                     AttestationTypes = new ushort[]
                     {
-                        15888
+                        (ushort)MetadataAttestationType.ATTESTATION_BASIC_FULL
                     },
                     Hash = "",
                     Description = "Windows Hello Hardware Authenticator",


aseigler:
I'll have a look at this as soon as I can.  I thought we had removed all the static metadata for authenticators that are now in MDS, which includes all of the ones in the static file...

m-gug:
Guess this is the result of an unfortunate coincidence of two problems. 
The one that the MDS implementation is currently not working because of the certificate change which triggers the fallback to the static metadata repo that has the wrong value in it.
In general i like the idea of having a fallback to the static repo. If all the entries that are now in MDS would have been removed from the static repo, it would be much bigger problem with the current MDS outage, wouldn't it?

aseigler:
> Guess this is the result of an unfortunate coincidence of two problems.
> The one that the MDS implementation is currently not working because of the certificate change which triggers the fallback to the static metadata repo that has the wrong value in it.

Yeah, I really am not impressed with that sudden root certificate change, especially considering all the flaming hoops that have to be jumped through to validate signatures on everything, verify CRLs, etc.  This is not how PKI operations works!

> In general i like the idea of having a fallback to the static repo. If all the entries that are now in MDS would have been removed from the static repo, it would be much bigger problem with the current MDS outage, wouldn't it?

Fallback and caching is great, it's actually required that a [server be able to "side load" its own custom metadata statements](https://github.com/abergs/fido2-net-lib/blob/master/Src/Fido2/Metadata/FileSystemMetadataRepository.cs) in order to pass conformance testing.  That would be my preferred fallback rather than having the library maintain a copy of the metadata statements.  The only reason the static metadata was included in the first place was because in years prior, even for the most popular devices like the ones that have been in the static metadata, the vendors for whatever reason didn't have their information published in the centralized and trusted source of information about FIDO authenticators, the [FIDO Alliance Metadata Service](https://fidoalliance.org/metadata/).  For that reason and others mentioned in the [FIDO MDS Purpose](https://fidoalliance.org/metadata/), we'll soon be removing the static metadata statements as redundant and/or outdated.

aseigler:
Working on this here https://github.com/abergs/fido2-net-lib/commit/2543e33cd0037c48bc9c49922db3350ed512063e

Issue #220, opened 2021-04-22
Checking Origin


DestruidorPT:
I was exploring the code, I have a question relative to the Origin, when an application sends the Authenticator Assertion Response, the origin of this response needs to be checked with the origin in the Authenticator Assertion request, correct?

Exemple:
- In the case of Web, a user requests to authenticate he has Origin “domain.com”, and then he sends a challenge signed with Origin “domain.com”, the origin of the request and the origin sent with the challenge signed must match.
- In the case of Android, a user requests to authenticate he has Origin “android:apk-key-hash:LBpmHljqwJJLdXVNpjfdA”, and then he sends a challenge signed with Origin “android:apk-key-hash:LBpmHljqwJJLdXVNpjfdA”, the origin of the request and the origin sent with the challenge signed must match.

So if the origin of the request must be the same as the origin sent with the challenge signed, it wouldn’t be better if instead of putting Origin in the Fido2Configuration, put in the function MakeAssertionAsync? This way Fido2Configuration doesn’t have to change every time a different origin comes, for example, the android sends a different origin from the web.

DestruidorPT:
My suggestion is this [DestruidorPT/fido2-net-lib](https://github.com/DestruidorPT/fido2-net-lib/commit/9d5cc3a425ceec34980cf883e9d35eaa9dd7d058), and I add a parameter on Fido2Configuration where we can configure what origins are allowed on FIDO2, this way if stop another app to access if the origin doesn't match, I also allowed to be empty in case someone wants the FIDO2 to be access from any app.

DestruidorPT:
I have corrected the missing origin in the test and in the demo, check my suggestion, if you agree I do a pull request.
[DestruidorPT/fido2-net-lib](https://github.com/DestruidorPT/fido2-net-lib)

Hinton:
We ran into the same issue while trying to add support for our Android application. The library is currently limited to a single origin which gets set during initialization.

Before I start looking into this in-depth, I wanted to check with the maintainers if you are open to having a community PR to add support for multiple origins?

Since we ideally want to maintain backwards compatibility with the current API, I think we can add a new optional parameter for `expectedOrigin` to the methods, which would take precedence over the `Fido2Configuration.Origin` value. Alternatively add a new setting `Origins` to the `Fido2Configuration` and fall back to `Origin` if it’s missing.

Does this seem like a reasonable approach @abergs? If so I would be happy to look into implement support for this.

[...]

PR #237, opened 2021-08-18
Support Multiple Origins

Hinton:
## Objective
Native android applications supports the FIDO2 protocol, but diverges slightly from the WebAuthN spec in that they set the `origin` field in the Attestation and Assertion to `android:apk-key-hash:<hash>`. Since setting `Fido2Configuration.Origin` to this value would cause web clients to fail we need a way to handle multiple allowed origins.

This PR deprecates the `Fido2Configuration.Origin` property and replace it with `Fido2Configuration.Origins`. It falls back gracefully to `Origin` should `Origins` not be set.

I aimed to not modifying any public API, but unfortunately both `AuthenticatorAssertionResponse` and `AuthenticatorResponse` signature was changed slightly to support a list of origins. Let me know if I should look into adding new methods instead to ensure backwards compatibility.

### Noteworthy Code Changes
- **Src/Fido2.Models/Fido2Configuration.cs**:
- **Src/Fido2/AuthenticatorAssertionResponse.cs**: Descriptions of file changes should be descriptive but brief, code should "speak for itself" however this enables the author to further describe the logic or theory behind the change.
- **Src/Fido2/AuthenticatorResponse.cs**: Changed `BaseVerify` to accept an list of origins.
- **Src/Fido2/Fido2NetLib.cs**: Changed to use `Origins` instead of `Origin`. This file seemed to contain mixed linebreaks. When saving they all changed to LF (`\n`). Let me know if we want to avoid these change)

Resolves: https://github.com/passwordless-lib/fido2-net-lib/issues/220


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/passwordless-lib/fido2-net-lib?pullRequest=237) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/passwordless-lib/fido2-net-lib?pullRequest=237) before we can accept your contribution.<br/><br/>:x: Hinton [sign now](https://cla.dotnetfoundation.org/passwordless-lib/fido2-net-lib?pullRequest=237)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/passwordless-lib/fido2-net-lib?pullRequest=237) it.</sub>

Src/Fido2/AuthenticatorResponse.cs
```diff
@@ -68,11 +69,11 @@ protected void BaseVerify(string expectedOrigin, byte[] originalChallenge, byte[
                 throw new Fido2VerificationException("Challenge not equal to original challenge");
 
             var fullyQualifiedOrigin = FullyQualifiedOrigin(Origin);
-            var fullyQualifiedExpectedOrigin = FullyQualifiedOrigin(expectedOrigin);
+            var fullyQualifiedExpectedOrigins = expectedOrigins.Select(FullyQualifiedOrigin);
 
             // 5. Verify that the value of C.origin matches the Relying Party's origin.
-            if (!string.Equals(fullyQualifiedOrigin, fullyQualifiedExpectedOrigin, StringComparison.OrdinalIgnoreCase))
-                throw new Fido2VerificationException($"Fully qualified origin {fullyQualifiedOrigin} of {Origin} not equal to fully qualified original origin {fullyQualifiedExpectedOrigin} of {expectedOrigin}");
+            if (!fullyQualifiedExpectedOrigins.Any(o => string.Equals(fullyQualifiedOrigin, o, StringComparison.OrdinalIgnoreCase)))
```
abergs, line 75: While not a dealbreaker, I'm wondering if a HashSet would be better than a List?
My gut feeling is that it would be able to do a lookup without iterating (like linq Any does).

It is mostly relevant if the list of origins grow due to large number of unique origins.

Src/Fido2/AuthenticatorResponse.cs
```diff
@@ -73,7 +73,7 @@ protected void BaseVerify(List<string> expectedOrigins, byte[] originalChallenge
 
             // 5. Verify that the value of C.origin matches the Relying Party's origin.
             if (!fullyQualifiedExpectedOrigins.Any(o => string.Equals(fullyQualifiedOrigin, o, StringComparison.OrdinalIgnoreCase)))
-                throw new Fido2VerificationException($"Fully qualified origin {fullyQualifiedOrigin} of {Origin} not equal to fully qualified original origin {fullyQualifiedExpectedOrigins} of {expectedOrigins}");
+                throw new Fido2VerificationException($"Fully qualified origin {fullyQualifiedOrigin} of {Origin} not equal to fully qualified original origin {string.Join(", ", fullyQualifiedExpectedOrigins)} of {string.Join(", ", expectedOrigins)}");
```
abergs, line 76: We should probably add a .Take(MAX_ORIGINS_TO_PRINT) to the list before string joining to stop a very long list of origins to generate huge strings in the exceptions

Hinton: Sure, any suggestion of what it should be set to?

Src/Fido2/AuthenticatorResponse.cs
Hinton: Sounds reasonable, just need to make sure to normalize the ordinal before comparing.

Src/Fido2/AuthenticatorResponse.cs
abergs: I think 5-10 should be enough.
Our original intention was to make sure the developer understod what value we were expecting. I think we could include a count just to avoid confusion now that we are hiding some.

Src/Fido2/AuthenticatorResponse.cs
Hinton: I changed it to HashSet, however to avoid iterating and converting them to FullyQualifiedOrigins each time, I had to store them in the `Fido2Configuration` object. I'm not completely happy with it, but I don't see any other place to put it.

aseigler:
So all the unit tests seem ok, but this PR breaks conformance tests currently.  This is what I am seeing:

```
Fido2NetLib.Fido2VerificationException: Fully qualified origin https://localhost:44329 of https://localhost:44329 not equal to fully qualified original origin  (1)
   at Fido2NetLib.AuthenticatorResponse.BaseVerify(HashSet`1 fullyQualifiedExpectedOrigins, Byte[] originalChallenge, Byte[] requestTokenBindingId) in source\repos\fido2-net-lib\Src\Fido2\AuthenticatorResponse.cs:line 81
   at Fido2NetLib.AuthenticatorAttestationResponse.VerifyAsync(CredentialCreateOptions originalOptions, Fido2Configuration config, IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser, IMetadataService metadataService, Byte[] requestTokenBindingId) in source\repos\fido2-net-lib\Src\Fido2\AuthenticatorAttestationResponse.cs:line 83
   at Fido2NetLib.Fido2.MakeNewCredentialAsync(AuthenticatorAttestationRawResponse attestationResponse, CredentialCreateOptions origChallenge, IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser, Byte[] requestTokenBindingId) in source\repos\fido2-net-lib\Src\Fido2\Fido2NetLib.cs:line 74
   at Fido2Demo.TestController.MakeCredentialResultTest(AuthenticatorAttestationRawResponse attestationResponse) in source\repos\fido2-net-lib\Demo\TestController.cs:line 106
 
```

Hinton:
@aseigler Hmm, I get a similar error if I leave `Origin` and `Origins` empty in `services.AddFido2`. I don't know much about the conformance process, but is it possible the `origin` is set to an empty value in `appsettings.json`?

We ended up not needing these changes after all, mainly due to iOS not natively supporting `WebAuthn` outside Safari (and `AsWebAuthenticationSession`), to minimize the dual maintenance we went with the same approach on Android as well. This allowed us to keep using only a single `Origin`. However I still believe there's value in getting this PR merged as it would add support for the native fido2 apis on Android.

aseigler (approved):

aseigler:
I am happy with this now. Thanks!

Hinton:
@aseigler Thanks for the help in getting this merged.

PR #241, opened 2021-10-07
Drop .NETStandard2.0 support on AspNet project

iamcarbon:
This PR simplifies the build targets and dependencies for the AspNet project and gets us closer to utilizing additional modern runtime features.

Changes:

- Removes netstandard2.0 target (This may break legacy .NET framework consumers)
- Removes netstandard2.1 target (We target the real runtimes directly)
- Adds a net5.0 target
- Uses a FrameworkReference to utilize the inbox assemblies
- Updates Xunit, Microsoft.NET.Test.Sdk, and coverlet dependencies


iamcarbon:
@aseigler Ready for feedback / review.

PR #245, opened 2021-10-08
Use .NET 5.0 SDK

iamcarbon:
Attempting to fix build.


iamcarbon:
@aseigler Ready for review. This just updates the SDK to 5.0.x and allows us to use .NET5.0 elsewhere. We're still targeting the same frameworks.

PR #247, opened 2021-10-08
Cleanup tests

iamcarbon:
This PR is focused on improving readability and performance of the test library. It also increases the framework target to .NET5.0.

This reduces the average test run time on my Surface Laptop from ~13s to ~8s.


iamcarbon:
@aseigler Ready for review.

PR #248, opened 2021-10-08
Remove require labels workflow

iamcarbon:
Contributors are not able to add labels and this workflow prevents us from going green.


iamcarbon:
GREEN!

aseigler:
I'm not familiar enough with the labeling stuff to be too much help here.  It would be nice if there was a way to require labels before a PR could be submitted, it's definitely not obvious during the submission process that you're supposed to provide a label.

abergs:
I think I read somewhere that Github have done some work on realeses. So they might have another way to automate release notes. Will take a look.

PR #249, opened 2021-10-09
Migrate from Newtonsoft to System.Text.Json

iamcarbon:
This PR migrates us from Newtonsoft to System.Text.Json. We should see reduced allocations & higher performance. 

We also added a fast-path for decoding UTF8-encoded Base64Url data that eliminates the extra string allocations.

The serializer is now case sensitive, which required a few fixes.


iamcarbon:
@aseigler Let me know if you have any other feedback. 

Once we merge, I'll start another round of optimization work / test coverage improvements - while we wait on the community response requiring .NET5.0.

aseigler:
The `FidoEnumConverter` doesn't seem to be working as expected, we're ending up with numerical values where we need strings. See https://github.com/passwordless-lib/fido2-net-lib/pull/204#issuecomment-810547911 and https://stackoverflow.com/questions/67129650/webauthn-credentials-create-exception.

PR #254, opened 2021-10-15
Improve code quality 4

iamcarbon:
This PR is pending the decision on bumping up the minimum framework to .NET5.0.

- Raise minimum framework target to net5.0
- Reduces allocations from hashing
- Eliminates System.Text.Json & System.Formats.Ans1 dependencies (they're inbox in .net5.0)
- Eliminates various other allocations
- Makes Asn1Element internal (exposed only to tests)
- Makes implementation helpers internal


aseigler:
Looking good to me.  Nobody has spoken up in opposition from .NET 5 either.

PR #261, opened 2021-11-03
Fix attestation certificate validation

adricasti:
The [attestationRootCertificates](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-metadata-statement-v2.0-id-20180227.html#widl-MetadataStatement-attestationRootCertificates) published in FIDO Metadata Services contain an array of trust anchors that are used to validate device certificates. It could happen that these trust anchors are themselves issued by some other root certification authority.  The .NET Framework X509Chain class correctly retrieves the untrusted root when building the chain, but the root is not published in MDS.  This patch fixes this case.


adricasti:
@abergs , @aseigler , if you need [Crescendo](https://www.hidglobal.com/product-display/cards-and-credentials/crescendo) devices to test or want to discuss more details about this patch, please let me know. You have put together a very nice library 👏

aseigler:
> @abergs , @aseigler , if you need [Crescendo](https://www.hidglobal.com/product-display/cards-and-credentials/crescendo) devices to test or want to discuss more details about this patch, please let me know. You have put together a very nice library 👏

Thanks!  I was going to ask or go research which devices fall into this category, as I believe none of the ones I currently have are set up like this.

aseigler:
Ended up entirely reworking ValidateTrustChain() using some new features in NET5 that allow us to more reliably build trust chains with full revocation checking support, and improving test coverage at the same time. Hopefully I can employ some similar techniques for other areas that have been plagued with related difficulties.

adricasti (commented):
I verified on my end as well and all works fine. This is a good improvement for the attestation chain validation. Thanks @aseigler

PR #266, opened 2021-11-14
Refactored IMetadataService to be async and removed init-related methods

mackie1001:
Refactored IMetadataService and implemented a new "on demand" implementation of DistributedCacheMetadataService that uses a combination of in-memory and distributed cache implementations.

Given that the initialisation mechanism for the service does not need to be exposed to the application or the rest of the library I've removed those.

Also, given the downstream impact of making the GetEntry method async was a single method that's already async (AuthenticatorAttestationResponse.VerifyAsync) it made sense to bin the old method.

The new DistributedCacheMetadataService implementation does not have to be a singleton (as it uses standard cache interfaces for its persistence) and does not require initialisation on startup (thus playing nice with things like Kubernetes and also the ability to self-heal should be the cache service be down when renewal is required).

Beefed up the tests a bit to check the caching boundary conditions.


mackie1001:
My only real question mark about this is the behaviour around expiry and actually keeping things in cache longer just in case the FIDO MDS is down when it comes to renewal and thus allowing a fallback. Could probably do with a second set of eyes and some additional test coverage.

aseigler:
> ...keeping things in cache longer just in case the FIDO MDS is down when it comes to renewal and thus allowing a fallback...

I am so much more comfortable with the MDS3 approach to this than I ever was with prior approaches, dare I say we don't necessarily need much of a fallback.

Src/Fido2/IMetadataService.cs
```diff
@@ -3,17 +3,14 @@
...
     public interface IMetadataService
     {
         /// <summary>
-        /// Gets the metadata payload entry by a guid.
+        /// Gets the metadata payload entry by a guid asyncronously
         /// </summary>
         /// <param name="aaguid">The Authenticator Attestation GUID.</param>
         /// <returns>Returns the entry; Otherwise <c>null</c>.</returns>
-        MetadataBLOBPayloadEntry? GetEntry(Guid aaguid);
+        Task<MetadataBLOBPayloadEntry?> GetEntryAsync(Guid aaguid);
```
abergs, line 13: My concern is just regarding optimization.
I have split opinions regarding this change:
I understand that keeping sync GetEntry is not really helping the library because the library code doesn't know if it the mds is optimized for sync/async. 

...But what I want to comment is:

> I don't think it is wise to remove the sync method.
> When calling the `SimpleMetdadataService` - the operation is in nature sync. Wrapping the result in a task causes overhead.
> 
> We should allow the caller to use either a sync method or async method on the interface, so *the caller* can choose what is optimal for their configuration (if they are using sync memetadataservice or async or custom)
> 
> For example, A warning or exception could be emitted from the `DistributedCacheMetadataService` if called sync.

...

But since *the caller* is the library, it has no idea to optmize for the sync mds. Any thoughts on this problem?

mackie1001: SimpleMetadataService is no longer used other than by ConformanceTesting.cs so I don't think it's an issue - we could probably get rid of it entirely now.
The framework caller is already async and thus has the overhead of Task/async state machine in play already as it already awaits another Task-returning method. 

I think the actual overhead in real terms is so slight and the benefits significant enough that "if it might involve some I/O make it return Task" can be a pretty hard and fast rule in .Net server application development even if the default/mock implementation (often memory-based in libs that provide such things - IdentityServer4 for example) doesn't actually need to do anything async.

abergs: Excellent. It should at least be properly renamed to avoid unwarranted usage.

mackie1001: I'll rename/remove it later today :)

mackie1001:
I've renamed the SimpleMetadataService to make the intent clearer

mackie1001:
@aseigler are you happy to include this in 3.0? let me know if there's anything more you want me take a look at.

mackie1001:
@aseigler  I've merged from upstream master, resolved the conflicts and incorportated the cancellation token changes into the new MDS stuff.

abergs:
Will review 👍

mackie1001:
Not to be impatient but do you have an idea of when this could be completed and 3.0 could be released? We have a project to upgrade our IDP to .Net 6 underway and it'd be nice to incorporate this upgrade too.

abergs:
Ok. I really want to get this done. Will find a way. I'm terribly sorry for the delay @mackie1001

Demo/appsettings.json
```diff
@@ -1,8 +1,8 @@
 ﻿{
   "fido2": {
     "serverDomain": "localhost",
-    "origin": "https://localhost:44329",
```
abergs, line 4: Why is this change made? Are we not running on https/localhost?

Src/Fido2/IMetadataService.cs
```diff
@@ -4,17 +4,14 @@
...
-        /// Gets the metadata payload entry by a guid.
+        /// Gets the metadata payload entry by a guid asyncronously
         /// </summary>
         /// <param name="aaguid">The Authenticator Attestation GUID.</param>
         /// <returns>Returns the entry; Otherwise <c>null</c>.</returns>
-        MetadataBLOBPayloadEntry? GetEntry(Guid aaguid);
+        Task<MetadataBLOBPayloadEntry?> GetEntryAsync(Guid aaguid, CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Gets a value indicating whether the internal access token is valid.
```
abergs, line 17: Is this comment really correct? I think it is wrong. 
I think ConformanceTesting() is used to inform the library if we're running in a conformance testing environment where "special rules" apply? @aseigler am I misstaken?

abergs (approved):
Great work. I approve the changes but I would like to see my comments replied before merging (at least the port change)

Demo/appsettings.json
abergs: I reverted the origin configuration to match the demo application.

abergs:
Never mind about waiting for replies. This PR has been stalled enough. I'm merging.

abergs:
Released on nuget as beta4: https://www.nuget.org/packages/Fido2/3.0.0-beta4

mackie1001:
Thanks for this, we'll pick it up shortly and give feedback :)

PR #270, opened 2021-12-04
Add .Net 6 as supported framework

trejjam:
Hi, there is already the .Net6 framework.
I prepared this small PR that adds it right next to the .Net5.

I found (and [resolved](https://github.com/trejjam/fido2-net-lib/commit/079b2e38f17cdab8be488e1f97c241d6681470e2)) one deprecation in packages: https://github.com/aspnet/Announcements/issues/303

### DIFF
--- a/Demo/.config/dotnet-tools.json
+++ b/Demo/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "dotnet-aspnet-codegenerator": {
+      "version": "6.0.0",
+      "commands": [
+        "dotnet-aspnet-codegenerator"
+      ]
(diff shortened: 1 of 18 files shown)


trejjam:
Do you mind if I add the PR check pipeline here to the Github Actions?

trejjam:
Hi, I think you are doing a lot of duplicity work in the pipelines. E.g. the prepare runtime is useless, since the next stages do a new checkout and as you can see the buld do a new restore.

Do you mind if I simplify it?

aseigler:
Go for it!

trejjam:
It should be ready

trejjam:
The pipeline time is now around half of the baseline.

aseigler:
Thanks!

PR #278, opened 2021-12-12
Add CancellationToken to Task methods

### LINKED ISSUE #271: Missing Task(CancellationToken)
Hello, I did notice that methods returning tasks do not accept the CancellationToken. Do you mind if I add it?

trejjam:
This PR adds `cancellationToken` parameters to Task methods.

Resolves #271

### DIFF
--- a/Demo/Controller.cs
+++ b/Demo/Controller.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Fido2NetLib;
 using Fido2NetLib.Development;
@@ -90,7 +91,7 @@ namespace Fido2Demo
 
         [HttpPost]
         [Route("/makeCredential")]
-        public async Task<JsonResult> MakeCredential([FromBody] AuthenticatorAttestationRawResponse attestationResponse)
+        public async Task<JsonResult> MakeCredential([FromBody] AuthenticatorAttestationRawResponse attestationResponse, CancellationToken cancellationToken)
         {
             try
             {
@@ -99,9 +100,9 @@ namespace Fido2Demo
                 var options = CredentialCreateOptions.FromJson(jsonOptions);
 
                 // 2. Create callback so that lib can verify credential id is unique to this user
-                IsCredentialIdUniqueToUserAsyncDelegate callback = async (IsCredentialIdUniqueToUserParams args) =>
+                IsCredentialIdUniqueToUserAsyncDelegate callback = static async (args, cancellationToken) =>
                 {
-                    var users = await DemoStorage.GetUsersByCredentialIdAsync(args.CredentialId);
+                    var users = await DemoStorage.GetUsersByCredentialIdAsync(args.CredentialId, cancellationToken);
                     if (users.Count > 0)
                         return false;
 
@@ -109,7 +110,7 @@ namespace Fido2Demo
                 };
 
                 // 2. Verify and make the credentials
-                var success = await _fido2.MakeNewCredentialAsync(attestationResponse, options, callback);
+                var success = await _fido2.MakeNewCredentialAsync(attestationResponse, options, callback, cancellationToken: cancellationToken);
 
                 // 3. Store the credentials in db
                 DemoStorage.AddCredentialToUser(options.User, new StoredCredential
@@ -177,7 +178,7 @@ namespace Fido2Demo
 
         [HttpPost]
         [Route("/makeAssertion")]
-        public async Task<JsonResult> MakeAssertion([FromBody] AuthenticatorAssertionRawResponse clientResponse)
+        public async Task<JsonResult> MakeAssertion([FromBody] AuthenticatorAssertionRawResponse clientResponse, CancellationToken cancellationToken)
         {
             try
             {
@@ -192,14 +193,14 @@ namespace Fido2Demo
                 var storedCounter = creds.SignatureCounter;
 
                 // 4. Create callback to check if userhandle owns the credentialId
-                IsUserHandleOwnerOfCredentialIdAsync callback = async (args) =>
+                IsUserHandleOwnerOfCredentialIdAsync callback = static async (args, cancellationToken) =>
                 {
-                    var storedCreds = await DemoStorage.GetCredentialsByUserHandleAsync(args.UserHandle);
+                    var storedCreds = await DemoStorage.GetCredentialsByUserHandleAsync(args.UserHandle, cancellationToken);
                     return storedCreds.Exists(c => c.Descriptor.Id.SequenceEqual(args.CredentialId));
                 };
 
                 // 5. Make the assertion
-                var res = await _fido2.MakeAssertionAsync(clientResponse, options, creds.PublicKey, storedCounter, callback);
+                var res = await _fido2.MakeAssertionAsync(clientResponse, options, creds.PublicKey, storedCounter, callback, cancellationToken: cancellationToken);
 
                 // 6. Store the updated counter
                 DemoStorage.UpdateCounter(res.CredentialId, res.Counter);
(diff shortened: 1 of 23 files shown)


Demo/Controller.cs
```diff
@@ -90,7 +91,7 @@ private string FormatException(Exception e)
 
         [HttpPost]
         [Route("/makeCredential")]
-        public async Task<JsonResult> MakeCredential([FromBody] AuthenticatorAttestationRawResponse attestationResponse)
+        public async Task<JsonResult> MakeCredential([FromBody] AuthenticatorAttestationRawResponse attestationResponse, CancellationToken cancellationToken)
```
iamcarbon, line 94: When is this cancellationToken called?

trejjam: When is HTTP request canceled (close tab, canceled XHR, ...)

aseigler: Having a cancellation token for a user action makes sense, not so sure about the MDS data load though, that should happen at startup and should not really be a cancellable event.

trejjam: Understood. I can remove it from here.

Anyway, a fired CancellationToken means no one is listening to the response. So it is meaningless to continue with request processing.

abergs:
Currently reviewing this 👀

Src/Fido2/SimpleMetadataService.cs
```diff
@@ -69,17 +70,17 @@ protected virtual async Task InitializeRepositoryAsync(IMetadataRepository repos
...
                         //Load if it doesn't already exist
-                        await LoadEntryStatementAsync(repository, blob, entry);
+                        await LoadEntryStatementAsync(repository, blob, entry, cancellationToken);
                     }
                 }
             }
         }
 
-        public virtual async Task InitializeAsync()
+        public virtual async Task InitializeAsync(CancellationToken cancellationToken)
```
abergs, line 79: Since this is a public api, the cancellation token should be optional.

Src/Fido2.AspNet/DistributedCacheMetadataService.cs
```diff
@@ -208,13 +210,13 @@ protected virtual async Task InitializeRepositoryAsync(IMetadataRepository repos
             }
         }
 
-        public virtual async Task InitializeAsync()
+        public virtual async Task InitializeAsync(CancellationToken cancellationToken)
```
abergs, line 213: I think we want CancellationToken to be optional here too

Demo/Controller.cs
abergs: @aseigler While not as usable, there are still situations that can occure that would benefit, at least in theory. Process wants to abort startup, e.g. fast restart, io hang etc. So best practice is to expose optional CancellationTokens in all async APIs.

Src/Fido2/Fido2NetLib.cs
```diff
@@ -64,10 +65,11 @@ public partial class Fido2 : IFido2
             AuthenticatorAttestationRawResponse attestationResponse,
             CredentialCreateOptions origChallenge,
             IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser,
-            byte[]? requestTokenBindingId = null)
+            byte[]? requestTokenBindingId,
+            CancellationToken cancellationToken)
```
abergs, line 69: Since this is an public API we want to make the cancellationToken optional

Src/Fido2/Fido2NetLib.cs
```diff
@@ -103,7 +105,8 @@ public partial class Fido2 : IFido2
             byte[] storedPublicKey,
             uint storedSignatureCounter,
             IsUserHandleOwnerOfCredentialIdAsync isUserHandleOwnerOfCredentialIdCallback,
-            byte[]? requestTokenBindingId = null)
+            byte[]? requestTokenBindingId,
+            CancellationToken cancellationToken)
```
abergs, line 109: Since this is an public API we want to make the CancellationToken optional.

Src/Fido2/Fido2NetLib.cs
```diff
@@ -140,11 +144,11 @@ public CredentialMakeResult(string status, string errorMessage, AttestationVerif
     /// </summary>
     /// <param name="credentialIdUserParams"></param>
     /// <returns></returns>
-    public delegate Task<bool> IsCredentialIdUniqueToUserAsyncDelegate(IsCredentialIdUniqueToUserParams credentialIdUserParams);
+    public delegate Task<bool> IsCredentialIdUniqueToUserAsyncDelegate(IsCredentialIdUniqueToUserParams credentialIdUserParams, CancellationToken cancellationToken);
```
abergs, line 147: Since this is an public API we want to make the CancellationToken optional.

Src/Fido2/Fido2NetLib.cs
```diff
@@ -140,11 +144,11 @@ public CredentialMakeResult(string status, string errorMessage, AttestationVerif
...
     /// <summary>
     /// Callback function used to validate that the Userhandle is indeed owned of the CrendetialId
     /// </summary>
     /// <param name="credentialIdUserHandleParams"></param>
     /// <returns></returns>
-    public delegate Task<bool> IsUserHandleOwnerOfCredentialIdAsync(IsUserHandleOwnerOfCredentialIdParams credentialIdUserHandleParams);
+    public delegate Task<bool> IsUserHandleOwnerOfCredentialIdAsync(IsUserHandleOwnerOfCredentialIdParams credentialIdUserHandleParams, CancellationToken cancellationToken);
```
abergs, line 153: Since this is an public API we want to make the CancellationToken optional.

Src/Fido2.AspNet/NullMetadataService.cs
```diff
@@ -15,7 +16,7 @@ MetadataBLOBPayloadEntry IMetadataService.GetEntry(Guid aaguid)
             return null;
         }
 
-        Task IMetadataService.InitializeAsync()
+        Task IMetadataService.InitializeAsync(CancellationToken cancellationToken)
```
abergs, line 19: Make token optional

abergs (changes requested):
I appreciateyour work! I want to make sure we make all the tokens optional for the public APIs.

I haven't looked through all the changes yet but that's what I found in my first read through.

Src/Fido2.AspNet/NullMetadataService.cs
trejjam: I think here it is not necessary. The method is in an internal class and the `default` is already defined on the interface. An external consumer cannot invoke the method directly.

Src/Fido2/Fido2NetLib.cs
trejjam: Delegates are public but without intention to be called by a consumer. Consumers need to define the new argument anyway. The gain of not specifying =default is a compile check for forgotten cancellationToken.

__
That is also the reason why I did not add =default to implementations in the first place. Consumers should mainly access through an interface (opinionated). And library implementation itself has a compile check when calling methods inside implementation (method A calling B in the same class).

Src/Fido2/Fido2NetLib.cs
trejjam: The same as https://github.com/passwordless-lib/fido2-net-lib/pull/278#discussion_r782596266

Src/Fido2/Fido2NetLib.cs
abergs: > Delegates are public but without intention to be called by a consumer.

Alright, I see your point on the delegates.

Src/Fido2.AspNet/NullMetadataService.cs
abergs: Ok

abergs:
LGTM 👍

abergs (approved):

abergs:
Great work @trejjam. Merged

PR #284, opened 2022-02-04
Allow unknown root when there is no Authority Information Access extension

adricasti:
When an intermediate certificate published to FIDO Metadata Server does not have Authority Information Access extension, we leave the flag to allow unknown root authority enabled so the validation of the leaf (device) certificate succeeds only with the intermediate certificate.

Maybe a cleaner option could be to check whether or not the intermediary certificate does contain the Authority Information Access extension (1.3.6.1.5.5.7.1.1) and if it does, then set the `chain.ChainPolicy.VerificationFlags = X509VerificationFlags.NoFlag;` but only for that case.

### DIFF
--- a/Src/Fido2/CryptoUtils.cs
+++ b/Src/Fido2/CryptoUtils.cs
@@ -89,21 +89,27 @@ namespace Fido2NetLib
             // try to build a chain with what we've got
             if (chain.Build(trustPath[0]))
             {
-                // if that validated, we should have a root for this chain now, add it to the custom trust store
-                chain.ChainPolicy.CustomTrustStore.Clear();
-                chain.ChainPolicy.CustomTrustStore.Add(chain.ChainElements[^1].Certificate);
+                // if that validated, and we have found a root for the chain, add it to the custom trust store
+                if (chain.ChainElements[^1].Certificate.Subject.Equals(chain.ChainElements[^1].Certificate.Issuer))
+                {
+                    chain.ChainPolicy.CustomTrustStore.Clear();
+                    chain.ChainPolicy.CustomTrustStore.Add(chain.ChainElements[^1].Certificate);
 
-                // explicitly trust the custom root we just added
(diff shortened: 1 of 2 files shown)


aseigler:
The only authenticator I see currently in MDS that doesn't have AIA (or CDP for that matter) is the `Crayonic KeyVault K1` and it is a root, not an intermediate.  Is there an example for this case somewhere?

adricasti:
Hi @aseigler, I just sent you a leaf certificate so you can test. The root of the problem is that the W3C WebAuthn specification recommends a [3-tier hierarachy](https://www.w3.org/TR/webauthn-2/#sctn-cert-hierarchy) for attestation certificates but the FIDO MDS specifies that [attestationRootCertificates](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-metadata-statement-v2.0-id-20180227.html#widl-MetadataStatement-attestationRootCertificates) are trust anchor for device certificates, so there is no expectation for full chain validation.  However, toolchains like the .NET `System.Security.Cryptography.X509Certificates` do automatic heavy lifting and retrieves the missing parts of the chain if AIA is present and accessible.

aseigler:
> The root of the problem is that the W3C WebAuthn specification recommends a [3-tier hierarachy](https://www.w3.org/TR/webauthn-2/#sctn-cert-hierarchy) for attestation certificates but the FIDO MDS specifies that [attestationRootCertificates](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-metadata-statement-v2.0-id-20180227.html#widl-MetadataStatement-attestationRootCertificates) are trust anchor for device certificates, so there is no expectation for full chain validation.

Understood.  It's unfortunate that there aren't more strict requirements in this space, having one or two vendors do this stuff in slightly different ways complicates things unnecessarily.

adricasti:
@aseigler we added Authority Information Access to the intermediary attestation _root_ so this change is no longer needed. I'm closing this pull request and maybe in the future we can think of ways to manage edge cases like this. Thanks again for your support.

PR #289, opened 2022-03-06
change：js code

hyfree:
js code
AttestationObject-->attestationObject
clientDataJson-->clientDataJSON

image


hyfree:
The propertyname in js should be the same as the propertyname of the c# code

PR #297, opened 2022-05-16
Add TLS endpoint to Demo project launch config

tmarkovski:
Adds both launch URLs to the Demo project config - this issue mostly affects Linux and MacOS users that don't have IIS Express configuration.
This allows the demo to be runnable and usable directly, instead throwing error where the HTTP url and port don't match the configured one

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -331,3 +331,4 @@ ASALocalRun/
 /results
 /Test/coverage.netcoreapp3.1.cobertura.xml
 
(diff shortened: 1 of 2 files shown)


abergs (approved):
LGTM. I'm on windows, so the only thing I was testing for was that settings did not change the existing behavior 👍

abergs:
Thanks for the contribution @tmarkovski

PR #302, opened 2022-05-27
Expose attestation certificates in attestation response.

### LINKED ISSUE #299: Provide Attestation Certificates for supported authenticators
When calling `MakeNewCredentialAsync`, the attestation certificate (if supported by the token), is not provided in the result.

This does make it impossible to store and/or examine the certificate of an authenticator afterwards, even if using the integrated metadata service.

boomer41:
Expose the optional attestation certificates in the attestation response to store and/or inspect them further.

Fixes #299

### DIFF
--- a/Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
+++ b/Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json.Serialization;
+﻿using System.Security.Cryptography.X509Certificates;
+using System.Text.Json.Serialization;
 
 namespace Fido2NetLib.Objects
 {
@@ -13,5 +14,7 @@ namespace Fido2NetLib.Objects
         public Fido2User User { get; set; }
         public string CredType { get; set; }
         public System.Guid Aaguid { get; set; }
+        public X509Certificate2 AttestationCertificate { get; set; }
+        public X509Certificate2[] AttestationCertificateChain { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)


Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
```diff
@@ -13,5 +14,7 @@ public class AttestationVerificationSuccess : AssertionVerificationResult
         public Fido2User User { get; set; }
         public string CredType { get; set; }
         public System.Guid Aaguid { get; set; }
+        public X509Certificate2 AttestationCertificate { get; set; }
```
iamcarbon, line 17: Add nullability annotations.

```csharp
public X509Certificate2? AttestationCertificate
```

boomer41: Will do, didn't see it's already used in this project. :)

Do you like it force-pushed in a single commit or as an additional commit?

Thanks
- Stephan

boomer41: Interestingly, adding the nullability annotation yields a compile error in the default solution.
`AttestationVerificationSuccess.cs(17, 32): [CS8632] The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.`

iamcarbon: This should work:

```csharp
#nullable enable
public X509Certificate2? AttestationCertificate { get; set; }
#nullable disable
```

boomer41: Of course that would work, however it feels like a big hack...
Will add for now, but maybe the nullability should be enabled on the whole project?

abergs: @boomer41 yes, nullable will eventually be turned on but we don't have the time to refactor the entire project to support it. So we do it piecewise.

boomer41:
Everything should be ready by now.

boomer41:
Thanks for merging @aseigler !
Are there already plans on the release of the next version?
Thanks!

PR #304, opened 2022-06-08
Use JSON source generator & enable trimming

iamcarbon:
This PR utilizes the JSON source generator and enables trimming. This should get us closer to enabling support for AOT builds.


iamcarbon:
@abergs @aseigler Ready for review.

Issue #310, opened 2022-07-25
Implement ResidenKeyRequirement


dIeGoLi:
When creating [CredentialCreateOptions](https://github.com/passwordless-lib/fido2-net-lib/blob/ef0501e016faaa8d4ec8137c6aba05aeb6fe8d72/Src/Fido2.Models/CredentialCreateOptions.cs#L72) we have the possibility to set `RequireResidentKey`. I observed that when using false, resident credentials are not created, thus usernameless flow is not possible. When set to true i tested on Android Firefox, that registration can be completed and requireResidentKey seems to be ignored, usernameless flow does not work. On android chrome following exception is thrown `DOMException: Either the device has received unexpected request parameters, or the device cannot support this request.`

W3C standard specifies for [Authenticator Selection Criteria](https://www.w3.org/TR/webauthn-2/#dictdef-authenticatorselectioncriteria) that there is a parameter `residentKey` of type [ResidentKeyRequirement](https://www.w3.org/TR/webauthn-2/#enum-residentKeyRequirement) with the values discouraged, preferred and required.

So i would like to extend the class [AuthenticatorSelection](https://github.com/passwordless-lib/fido2-net-lib/blob/ef0501e016faaa8d4ec8137c6aba05aeb6fe8d72/Src/Fido2.Models/CredentialCreateOptions.cs#L184) with a property `ResidentKeyRequirement ResidentKeyRequirement {get; set; } = ResidentKeyRequirement.Required` to enable using this further specification. I already tested the behaviour with `residentKey: "preferred"` and it works as expected. If resident key is supported it registers as resident key, otherwise not.

Would you accept a pull request regarding this?

aseigler:
This looks really familiar...I thought I had run into and fixed this on branch [demo-update](https://github.com/passwordless-lib/fido2-net-lib/tree/demo-update), but I don't see it currently.  We would certainly accept a PR here.

abergs:
Yes, I agree.

reitowo:
Is this even fixed?

PR #277, opened 2021-12-06
(WIP) Demo site update

aseigler:
- Improvements mostly to custom area of demo, lots more options (inspired largely by https://webauthntest.azurewebsites.net)
- Default everywhere to UV = discouraged
- Integrate info from MDS when AAGUID is available
- Prioritize Ed25519 when available

PR #311, opened 2022-07-25
Support ResidentKeyRequirement

dIeGoLi:
#310 
I can not think of a good unit test. I think it does not need one, right?


abergs (approved):
I think this looks good.

dIeGoLi:
@aseigler Would be nice to have a v3 beta release with that. Do you think that's possible?

abergs:
I could probably get this into a 3.1 beta

Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -205,6 +212,7 @@ public class AuthenticatorSelection
         public static AuthenticatorSelection Default => new AuthenticatorSelection
         {
             AuthenticatorAttachment = null,
+            ResidentKey = ResidentKeyRequirement.Required,
```
aseigler, line 215: I'm good with the resident key requirement enum, but why do we want the default to be required?  This feels problematic.

dIeGoLi: I chose it to be Required based on following sentence in the [standard](https://w3c.github.io/webauthn/#dictionary-authenticatorSelection):

> **If no value is given then the effective value is [required](https://w3c.github.io/webauthn/#dom-residentkeyrequirement-required) if [requireResidentKey](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-requireresidentkey) is true** or [discouraged](https://w3c.github.io/webauthn/#dom-residentkeyrequirement-discouraged) if it is false or absent.

But my interpretation was wrong. The default value should be `discouraged` because `requireResidentKey` is default `false` and an absent `residentKey` property would then be defaulted to `discouraged`.

If I think further about the connection these two properties have, then changing one should change the other. For example it makes no sense for a user of the library to set `requireResidentKey=true and residentKey=discouraged`
I see two possible approaches.

Non breaking approach:
- Set requireResidentKey to true sets residentKey to required
- Set requireResidentKey to false sets residentKey to discouraged
- Set residentKey to discouraged sets requireResidentKey to discouraged
- Set residentKey to preferred sets requireResidentKey to discouraged (requireResidentKey: [Relying Parties](https://w3c.github.io/webauthn/#relying-party) SHOULD set it to true if, and only if, [residentKey](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-residentkey) is set to [required](https://w3c.github.io/webauthn/#dom-residentkeyrequirement-required).)
- Set residentKey to required sets requireResidentKey to true

Breaking approach but better because of no redundant properties:
- Set property `RequireResidentKey` to private and apply the same mapping as specified above.
- Set residentKey to discouraged sets requireResidentKey to discouraged
- Set residentKey to preferred sets requireResidentKey to discouraged (requireResidentKey: [Relying Parties](https://w3c.github.io/webauthn/#relying-party) SHOULD set it to true if, and only if, [residentKey](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-residentkey) is set to [required](https://w3c.github.io/webauthn/#dom-residentkeyrequirement-required).)
- Set residentKey to required sets requireResidentKey to true

Which option would you prefer that i implement @aseigler?

dIeGoLi: What I am not sure is, if we could remove requireResidentKey completely.
About requireResidentKey: 

> This member is retained for backwards compatibility with WebAuthn Level 1...

Does the specification describe it for backward compatibility or should the RP include it for backward compatibility...

dIeGoLi: I am struggling to follow you on the the pr 227. Could it be that you linked the wrong one?
I checked the webauthntest and saw that they allow choosing "Require Resident Key" and "Resident Key (L2)". I think as well that it is elegant to include only selected options using the `!==` operator in js. But I don't think that it is useful to include both properties for the demo in general. Require Resident Key is retained for backwards compatibility and is specified definite through ResidentKey. I see no functional need to include both. It only opens the possibility to choose incompatible values. Could you give me a clearer hint what you would do different?

aseigler: Oof, yes, I meant #277. I agree backward compatibility doesn't really make any sense to keep for the demo for this property, just makes things more confusing.

dIeGoLi: Oh ok. I see that you included requireResidentKey and residentKey in the demo in this branch. But you didn't integrate it into the c# code. The second property (residentKey) is set in the javascript code to `makeCredentialOptions.authenticatorSelection` after the makeCredentialOptions have been fetched from the server.

I see two options.
I integrate the code of this pull request into 277, but i will remove the `requireResidentKey` option, and i will keep the c# integration, so that in all the c# and javascript code only `residentKey` is used.
Or we merge this pullrequest into the master and then i will merge the master into your demo branch and do the same clean up. What would suit you better? What is the timeline for the demo branch? :)

aseigler:
Works for me, sorry it took so long to sort.

PR #312, opened 2022-07-26
IOptions changed to IOptionsMonitor for better options handling.

se:
For better cases like multi-tenancy, options need to change after the app is started. We can change the IOptions to IOptionsMonitor to handle this case.

I have updated Main `Fido2` class, and all Test Classes contain `Fido2`.

More explanation in here; https://docs.microsoft.com/en-us/dotnet/core/extensions/options#:~:text=Supports%20named%20options-,IOptionsMonitor%3CTOptions%3E%3A,-Is%20used%20to

### DIFF
--- a/Demo/TestController.cs
+++ b/Demo/TestController.cs
@@ -24,13 +24,13 @@ namespace Fido2Demo
         public TestController(IOptions<Fido2Configuration> fido2Configuration)
         {
             _origin = fido2Configuration.Value.FullyQualifiedOrigins.FirstOrDefault();
-
-            _fido2 = new Fido2(new Fido2Configuration
+            
+            _fido2 = new Fido2(new TestOptionsMonitor<Fido2Configuration>(new Fido2Configuration
             {
                 ServerDomain = fido2Configuration.Value.ServerDomain,
                 ServerName = fido2Configuration.Value.ServerName,
                 Origins = fido2Configuration.Value.FullyQualifiedOrigins,
-            }, 
+            }), 
             ConformanceTesting.MetadataServiceInstance(
                 System.IO.Path.Combine(fido2Configuration.Value.MDSCacheDirPath, @"Conformance"), _origin)
             );
--- a/Demo/TestOptionsMonitor.cs
+++ b/Demo/TestOptionsMonitor.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using Microsoft.Extensions.Options;
+
+namespace Fido2Demo
+{
+    public class TestOptionsMonitor<TOptions> : IOptionsMonitor<TOptions>
+    {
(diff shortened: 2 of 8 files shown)


abergs (changes requested):
Hmm. I appreciate your thought of contribution but have some concerns.
I do not like the removal of Fido2Configuration in the Fido2 class and introduction of IOptionsMonitor. I think it complicates the code base.

What exactly is the use case for this?
For multi-tenancy (where options differ), you would need separate instances of the Fido2 library to make sure you are not changing options in the middle of an authentication.

se:
Let's say you have 3 different hostnames on the same app. Those hostnames define the tenant of your customers.

For example;
`selcuk.passwordless.dev` (This is customer Selcuk)
`anders.passwordless.dev` (This is customer Anders)
`another.passwordless.dev` (This is customer Another)

If you handle your Fido class as a singleton class or Options class as singleton, you can't change that config while your action is executing. That is the actual case of this. Even if there is no case like this, it is better to separate your base class or the way you handle the options.

In my opinion, it is better to change the way I did. Users will not be affected. Everything will be handled by the DI. 

I hope I did make it clear. If you need an example project, just let me know, and I will make it (even though it is time-consuming 😃).

abergs:
I'm the case of multi tenancy you could not have a singleton Fido2 class and modify the options during runtime. If two tenants perform Auth at the same time there is a high probability that it would fail due to options being swapped half-way through the process.

The fido2 class is not written to be used like that, and proxying the config would not change that.

My recommendation for your use case is to re-create and instance of Fido2 with the updated or changed options.

se:
That is why IOptionsMonitor has been made. As I mentioned before, the case is the same with Microsoft Documentation. Even though this is Dependency Injection lifecycle that Microsoft offers you for more extensibility. 

CleanShot 2022-08-01 at 08 37 45@2x

Decision is yours 🖖

abergs:
I appreciate your intentions and effort, but I still believe your solution is flawed.
This quote matches my understanding:

> IOptionsMonitor is a singleton service that retrieves current option values at any time, which is especially useful in singleton dependencies.

(from: https://stackoverflow.com/questions/50788988/difference-between-ioptionsmonitor-vs-ioptionssnapshot)

* Fido2 should NOT be run as a singleton instance TO to support multi-tenancy via different options. It will break. Instead you should instantiate multiple different Fido2 instances with each tenant options.
* Fido2 CAN BE RUN as a singleton, but then you should not change your config values.
* If you want to change configuration, create a new instance of Fido2

In theory, your solution could be useful in a scenario where you really need/want to have Fido2 as a singleton but be able to update the configuration. But, that's only in theory, you would be better off to just re-instantiate your Fido2-class.

Closing, but appreciating your effort to contribute.

PR #322, opened 2022-09-06
Implement support for apple-appattest attestation format

### LINKED ISSUE #319: Feedback on using library for Apple App Attest
I've recently had to implement app validation on a REST API for Apple's App Attest Service (which is based around issuing a WebAuthn Attestation object). Apple's documentation is located at https://developer.apple.com/documentation/devicecheck/establishing_your_app_s_integrity (the WebAuthn Attestation specific stuff is at https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server, which was previously mentioned in #184).

Unfortunately, due to reasons I'll detail below, the primary classes from Fido2NetLib (mainly `AuthenticatorAttestationResponse.VerifyAsync()`) can't be entirely used in this scenario. So I wanted to offer this feedback/open a discussion with the aim of being able to use Fido2NetLib entirely in some future update.

1. The main roadblock is that the `AttestationObject.Fmt` for Apple's App Attest Service has a value of `apple-appattest`. Currently there's no way to inject custom implementations of `AttestationVerifier`, so we'll get a `Fido2VerificationException` thrown from `AuthenticatorAttestationResponse.VerifyAsync()`. (A hack-around by changing `ParsedAttestationObject.Fmt` isn't possible as this property doesn't have a publicly accessible setter). In #184 mackie1001 suggested having attestation formats pluggable (and aseigler mentioned structural refactoring around this for v2.0), however I'm not sure if that was heading to an injectable `AttestationVerifier` factory or resolver that would be needed in this situation. (If it isn't and nobody else is working on one, I could see about submitting a PR for this if I can find the time).

2. Another difficulty is that `apple-appattest` (and the corresponding root CA) isn't listed in the spec's Defined Attestation Statement Formats. Because of 1. I didn't continue investigating this. However, would I be correct that supplying a custom implementation of `IMetadataService` to `AuthenticatorAttestationResponse.VerifyAsync()` that would provide the Apple App Attest Root CA from https://www.apple.com/certificateauthority/private/ would be all that is needed to complete the validation of certificates supplied through the attestation statement?

3. Steps 5 and 9 of Apple's documentation mentions to verify the `credentialId` and SHA256 hash of the `credCert` public key against the `keyId` (obtained from `DCAppAttestService.shared.generateKey()`, and supplied to `DCAppAttestService.shared.attestKey()` to generate the attestation object). The existing `Apple` implementation of `AttestestationVerifier` compares those values from the attestation object itself, but doesn't provide a way to check them against some externally provided value. I suppose this could be passed in a delimited `clientDataJson`, but points to extend the current classes to achieve this don't currently exist (going back to 1.)

I'm unsure if there might be any similar potential difficulties around the assertion objects as I'm not currently using them or the receipts mentioned in the Apple documentation in our REST API, but likely will at some point in the future.

Thanks for what looks like a fantastic library at the moment (and certainly the premier library for working with WebAuthn objects in .NET). I appreciate any advice or concerns about the above points you all may have.

aseigler:
Resolves #319


Src/Fido2/AttestationFormat/AppleAppAttest.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            // 1. Verify that the x5c array contains the intermediate and leaf certificates for App Attest, starting from the credential certificate in the first data buffer in the array (credcert).
+            if (!(X5c is CborArray { Length: 2 } x5cArray && x5cArray[0] is CborByteString { Length: > 0 } && x5cArray[1] is CborByteString { Length: > 0 }))
+            {
+                throw new Fido2VerificationException("Malformed x5c in Apple AppAttest attestation");
+            }
+
+            // Verify the validity of the certificates using Apple's App Attest root certificate.
+            var chain = new X509Chain();
+            chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
```
gevoulga, line 59: Maybe: NoRevocationCheck when in dev environment?

Src/Fido2/AttestationFormat/AppleAppAttest.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            // 7. Verify that the authenticator data's counter field equals 0.
+            if (AuthData.SignCount != 0)
+            {
+                throw new Fido2VerificationException("Sign count does not equal 0 in Apple AppAttest attestation");
+            }
+
+            // 8. Verify that the authenticator data's aaguid field is either appattestdevelop if operating in the development environment, or appattest followed by seven 0x00 bytes if operating in the production environment.
+            if (!AuthData.AttestedCredentialData.AaGuid.Equals(devAaguid) && !AuthData.AttestedCredentialData.AaGuid.Equals(prodAaguid))
```
gevoulga, line 110: Maybe the apple env should be a config that we validate?
ie. expecting from dev-> verify Aaguid matches dev

Src/Fido2/AttestationFormat/AppleAppAttest.cs
```diff
@@ -0,0 +1,124 @@
...
+                        return s[0].GetOctetString();
+                    }
+                }
+            throw new Fido2VerificationException("Apple AppAttest attestation extension 1.2.840.113635.100.8.5 has invalid data");
+        }
+
+        // From https://www.apple.com/certificateauthority/Apple_App_Attestation_Root_CA.pem
+        internal static readonly string appleAppAttestationRootCA = "MIICITCCAaegAwIBAgIQC/O+DvHN0uD7jG5yH2IXmDAKBggqhkjOPQQDAzBSMSYwJAYDVQQDDB1BcHBsZSBBcHAgQXR0ZXN0YXRpb24gUm9vdCBDQTETMBEGA1UECgwKQXBwbGUgSW5jLjETMBEGA1UECAwKQ2FsaWZvcm5pYTAeFw0yMDAzMTgxODMyNTNaFw00NTAzMTUwMDAwMDBaMFIxJjAkBgNVBAMMHUFwcGxlIEFwcCBBdHRlc3RhdGlvbiBSb290IENBMRMwEQYDVQQKDApBcHBsZSBJbmMuMRMwEQYDVQQIDApDYWxpZm9ybmlhMHYwEAYHKoZIzj0CAQYFK4EEACIDYgAERTHhmLW07ATaFQIEVwTtT4dyctdhNbJhFs/Ii2FdCgAHGbpphY3+d8qjuDngIN3WVhQUBHAoMeQ/cLiP1sOUtgjqK9auYen1mMEvRq9Sk3Jm5X8U62H+xTD3FE9TgS41o0IwQDAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBSskRBTM72+aEH/pwyp5frq5eWKoTAOBgNVHQ8BAf8EBAMCAQYwCgYIKoZIzj0EAwMDaAAwZQIwQgFGnByvsiVbpTKwSga0kP0e8EeDS4+sQmTvb7vn53O5+FRXgeLhpJ06ysC5PrOyAjEAp5U4xDgEgllF7En3VcE3iexZZtKeYnpqtijVoyFraWVIyd/dganmrduC1bmTBGwD";
```
gevoulga, line 36: maybe get the RootCA from the url (and use a hardwired one as a fallback?)
It's unlikely, but I am thinking of the case where it's revoked.

aseigler: Hard coding the root is the best practice. You can't revoke a root CA certificate.

Src/Fido2/AttestationFormat/AppleAppAttest.cs
aseigler: In the samples I have seen for both dev and prod there are no CDP extensions present, so no way to check CRL regardless.

Test/Attestation/AppleAppAttest.cs
```diff
@@ -0,0 +1,152 @@
...
+        var cbor = Convert.FromBase64String(b64);
+        _attestationObject = (CborMap)CborObject.Decode(cbor);
+        _aaguid = new("61707061-7474-6573-7464-6576656c6f70");
+
+        //_attestationObject = new ParsedAttestationObject
+        //(
+        //    fmt: (string)json["fmt"],
+        //    attStmt: (CborMap)json["attStmt"],
```
gevoulga, line 28: What about the `receipt` which is part of attStmt?
This will be needed if the counter field needs to be verified later:
https://developer.apple.com/documentation/devicecheck/assessing_fraud_risk?language=objc

aseigler: I don't know what people do with the `receipt` field and it's not in the WebAuthn spec so I was just ignoring that for now.  The `counter` is part of the `authData` field and is handled just like any other attestation format.

Src/Fido2/AttestationFormat/AppleAppAttest.cs
aseigler: I can see where that could be useful if you wanted to not accept dev devices in a production only environment.

Src/Fido2/AuthenticatorAttestationResponse.cs
```diff
@@ -145,6 +145,7 @@ public static AuthenticatorAttestationResponse Parse(AuthenticatorAttestationRaw
                 "fido-u2f" => new FidoU2f(),                    // https://www.w3.org/TR/webauthn/#fido-u2f-attestation
                 "packed" => new Packed(),                       // https://www.w3.org/TR/webauthn/#packed-attestation
                 "apple" => new Apple(),                         // https://www.w3.org/TR/webauthn/#apple-anonymous-attestation
+                "apple-appattest" => new AppleAppAttest(),      // https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server
```
jrloe, line 148: This comment is not related to this line, but the user present check on line 118.
Do we know if Apple App Attest provides a user presence check?

aseigler: Doesn't much matter to me, but if it doesn't it's not compliant with [WebAuthn spec](https://w3c.github.io/webauthn/#sctn-registering-a-new-credential), step 13.

PR #330, opened 2022-10-06
added [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] annotations to generic enums

filipw:
At the moment the build fails when using .NET SDK `7.0.100-rc.1`
There is a trimming error:

> /fido2-net-lib/Src/Fido2.Models/Converters/EnumNameMapper.cs(68,35): error IL2090: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to 'System.Type.GetFields(BindingFlags)'. The generic parameter 'TEnum' of 'Fido2NetLib.EnumNameMapper<TEnum>' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/Users/filipw/Documents/dev/fido2-net-lib/Src/Fido2.Models/Fido2.Models.csproj::TargetFramework=net6.0]

This PR adds the necessary annotations.

### DIFF
--- a/Src/Fido2.Models/Converters/EnumNameMapper.cs
+++ b/Src/Fido2.Models/Converters/EnumNameMapper.cs
@@ -1,11 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Runtime.Serialization;
 
 namespace Fido2NetLib
 {
-    public static class EnumNameMapper<TEnum>
+    public static class EnumNameMapper<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] TEnum>
         where TEnum: struct, Enum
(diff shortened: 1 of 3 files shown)


abergs:
Thank you @filipw. Looks good, will do a quick review and build/test on my machine before merging.

abergs (approved):

abergs:
@filipw Every looks good. Merging. Where can I read more about the changes in .net 7 that causes these trimming issues?

filipw:
There is some documentation around what to do in the library to prepare it for trimming here https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming and it includes some info around these attributes. These are good practices to follow anyhow, but I am not sure why it suddenly showed up as error with .NET 7 SDK 🤷‍♂️

PR #332, opened 2022-10-12
Add Mac and Linux to test pipeline

aseigler:
Historically there have been several times where we've discovered code changes or test additions that work just fine on Windows, but are broken on Mac or Linux.  This PR should help proactively fill that gap by:

- Marking existing demo/test/coverage jobs to make it clear they are Windows specific
- Adding job to test Mac and Linux
- Updating tests where there are known/expected runtime issues


abergs (approved):
YES @aseigler!

aseigler:
The Mac and Linux tests both have failing tests outstanding, but this should help bring some visibility.

abergs:
@aseigler Something weird on this one, we don't run the tests. Investigating...

PR #341, opened 2022-11-18
Doing both length 0 and containskey checks for x5c subject

### LINKED ISSUE #340: Packed.cs assumes x5c subject contains O, OU, C and CN
The code in the `IsValidPackedAttnCertSubject` of the [Packed.cs ](https://github.com/passwordless-lib/fido2-net-lib/blob/master/Src/Fido2/AttestationFormat/Packed.cs) uses the `Item[TKey]` method from `Dictionary` on the keys O, OU, C and CN. If the x5c doesn't contain one of these keys it will result in an `KeyNotFoundException`. The `KeyNotFoundException` is then not caught resulting in the `Verify` method not throwing the more appropriate `Fido2VerificationException`

Molyna:
Fixes #340 

Made the code check that the subject field exists before checking the length of it to generate the a better exception being thrown by the verify code.


abergs:
LGTM! Anything you want to comment on @aseigler?

Molyna:
I'll close this then as it has been overwritten by #344 which  has the same functionality.

PR #344, opened 2022-11-24
Improve error handling

iamcarbon:
In addition to addressing #340 (and superseding #341) to address the merge conflict:

This PR:
- Improves attnCert property validation and eliminates extra dictionary lookups
- Moves more error messages to Fido2ErrorMessages
- Eliminates unneeded LINQ casts [.NET added strongly typed enumerators]
- [BREAKING] Makes AttestationVerifier{attStmt,authenticatorData,clientDataHash} private protected.  If needed by consumers, we should expose over public properties.


iamcarbon:
@abergs @aseigler Ready for review / feedback.

Please note the field visibility changes made in AttestationVerifier and thoughts on whether we should expose new public properties.

aseigler:
> @abergs @aseigler Ready for review / feedback.
> 
> Please note the field visibility changes made in AttestationVerifier and thoughts on whether we should expose new public properties.

I was looking to add something like the below to AttestationVerifer and move the similar code out of the `AuthenticatorAttestationResponse.VerifyAsync()`:

```cs
public static AttestationVerifier Create(string fmt)
{
    return fmt switch
    {
        "none" => new None(),                           // https://www.w3.org/TR/webauthn/#none-attestation
        "tpm" => new Tpm(),                             // https://www.w3.org/TR/webauthn/#tpm-attestation
        "android-key" => new AndroidKey(),              // https://www.w3.org/TR/webauthn/#android-key-attestation
        "android-safetynet" => new AndroidSafetyNet(),  // https://www.w3.org/TR/webauthn/#android-safetynet-attestation
        "fido-u2f" => new FidoU2f(),                    // https://www.w3.org/TR/webauthn/#fido-u2f-attestation
        "packed" => new Packed(),                       // https://www.w3.org/TR/webauthn/#packed-attestation
        "apple" => new Apple(),                         // https://www.w3.org/TR/webauthn/#apple-anonymous-attestation
        "apple-appattest" => new AppleAppAttest(),      // https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server 
        _ => throw new Fido2VerificationException(Fido2ErrorCode.UnknownAttestationType, $"Unknown attestation type. Was '{fmt}'")
    };
}
```

Then have the caller call `Create(fmt)` to get a verifier.  Right now this only gets called from inside the attestation verifier, but with the passkey support, it can get called when a devicePublicKey extension object is found in an attestation or inside an assertion.

Src/Fido2/AttestationFormat/AndroidKey.cs
```diff
@@ -134,21 +134,20 @@ public override (AttestationType, X509Certificate2[]) Verify()
...
         // 1. Verify that attStmt is valid CBOR conforming to the syntax defined above and perform CBOR decoding on it to extract the contained fields
         // (handled in base class)
-        if (attStmt.Count is 0)
-            throw new Fido2VerificationException(Fido2ErrorCode.InvalidAttestation, "Attestation format android-key must have attestation statement");
+        if (_attStmt.Count is 0)
+            throw new Fido2VerificationException(Fido2ErrorCode.InvalidAttestation, Fido2ErrorMessages.MissingAndroidKeyAttestationStatement);
 
-        if (!(Sig is CborByteString { Length: > 0 }))
-            throw new Fido2VerificationException(Fido2ErrorCode.InvalidAttestation, "Invalid android-key attestation signature");
+        if (!TryGetSig(out byte[]? sig))
```
abergs, line 140: It's unfortunate that this performance boost is a little less readable. Never been a fan of  `out`, but I think the trade off is OK, just unfortunate.

abergs (commented):
I've did a quick review of 12/13 files, which is fine.
I haven't looked at the big bulk of changes in Src/Fido2/AttestationFormat/AttestationVerifier.cs.

I like @aseigler approach.

iamcarbon:
@aseigler I went ahead and moved the logic for creating AttestationVerifier instances up. Much nicer.

aseigler:
> @aseigler I went ahead and moved the logic for creating AttestationVerifier instances up. Much nicer.

So glad you like it.  I have been staring at that area for the past....I want to say, 4 years, and keep thinking `"there's got to be a better way!"` but didn't really see it until the DPK work.  I will try my best to get this merged early this weekend.

abergs (commented):
Indeed, really nice.

aseigler (approved):

PR #351, opened 2022-12-05
Improve X509 Certificate construction / error handling

iamcarbon:
This PR:

- Consolidates the X509Certificate constructor / error handlers in X509CertificateHelper. We know throw an improved error message if there was invalid Base64 data.
- Uses the built-in JSON converter to decode base64 data. System.Text.Json can convert directly from the utf-8 bytes using the vectorized Base64 converter.


iamcarbon:
This history got messed up a bit here. I'll submit a clean PR later -- but this is ready for initial review / feedback.

Issue #293, opened 2022-04-27
Inconsistent chain validation on Windows / Linux which results to 2 failed unit tests on Linux


dbalikhin:
It works on Windows, but on Linux there is a different behavior that seems actually normal, see below.

Failed tests:
- CryptoUtilsTests.TestValidateTrustChainSubAnchor() line 66
- AuthenticatorAttestationResponse.VerifyAsync(CredentialCreateOptions originalOptions, Fido2Configuration config, IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser, IMetadataService metadataService, Byte[] requestTokenBindingId, CancellationToken cancellationToken) line 187
Fido2Tests.TestInvalidU2FAttestationASync() line 624

Message: 
Fido2NetLib.Fido2VerificationException : Invalid certificate chain

From:
https://github.com/passwordless-lib/fido2-net-lib/blob/db931bdba20aadc0743184fc9b0383a88b56ce02/Src/Fido2/CryptoUtils.cs#L107-L118

What happens in ValidateTrustChain:
1.             // try to build a chain with what we've got
            if (chain.Build(trustPath[0]))
We disable revocation flag and allow unknown CA

On Windows we will get success + Status Untrusted Root (A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.) which is expected.
On Linux we will get success + Status Partial Chain (Unable to find local issuer)

2.                 // now, verify chain again with all checks turned on
                if (chain.Build(trustPath[0]))
This one will succeed on Windows but fail on Linux. The required item won't be added to CustomTrustStore on Linux, because "chain.ChainElements[^1].Certificate" doesn't contain it.

So we will have something like this:

            if (chain.Build(trustPath[0])) // first chain build
            {
                if (chain.ChainStatus[0].Status == X509ChainStatusFlags.UntrustedRoot)
                {
                    // if that validated, we should have a root for this chain now, add it to the custom trust store
                    chain.ChainPolicy.CustomTrustStore.Clear();
                    chain.ChainPolicy.CustomTrustStore.Add(chain.ChainElements[^1].Certificate);
                    
                }
                else if (chain.ChainStatus[0].Status == X509ChainStatusFlags.PartialChain)
                {
                    ???  
                    Chain is validated but we won't have a root of this chain. 
                }
..
}

**Any ideas on how to handle PartialChain on Linux then?**

[...]

PR #355, opened 2022-12-12
Update tests to get clean CI

aseigler:
- Add port to conformance metadata repository - `http://localhost` no longer seems to work, `http://localhost:80` works fine
- Temporarily disable 2 specific tests that don't work on Mac/Linux


aseigler:
Related to #293

Issue #350, opened 2022-12-03
Feature request: Support for credProps extension


dbeinder:
When creating a credential with `ResidentKey` set to `preferred` or `discouraged` the RP needs the `credProps` extension to be able to tell whether the newly created credential is resident or not:
https://w3c.github.io/webauthn/#credprops

abergs:
Solived in #411

PR #356, opened 2022-12-13
WIP: Support device public key and passkeys

### LINKED ISSUE #334: Library lacks support for passkeys and devicePublicKey extension
This library currently lacks support for [multi-device FIDO credentials](https://w3c.github.io/webauthn/#backup-eligibility) (aka [passkeys](https://passkeys.dev/)) as well as the related [devicePublicKey ](https://w3c.github.io/webauthn/#sctn-device-publickey-extension)extension which lets passkeys not be a deal-breaker for [enterprise](https://w3c.github.io/webauthn/#dom-attestationconveyancepreference-enterprise) customers.  I should fix this.

aseigler:
Resolves #334 (and #350).


Src/Fido2.Models/Objects/AuthenticationExtensionsDevicePublicKeyInputs.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Text.Json.Serialization;
+
+namespace Fido2NetLib.Objects
```
iamcarbon, line 4: Use a file scoped namespace?

Src/Fido2.Models/Objects/AuthenticationExtensionsDevicePublicKeyOutputs.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Fido2NetLib.Objects
```
iamcarbon, line 3: Use a file scoped namespace?

Src/Fido2/Objects/AuthenticatorData.cs
```diff
@@ -41,6 +41,22 @@ public sealed class AuthenticatorData
...
     public bool UserVerified => _flags.HasFlag(AuthenticatorFlags.UV);
 
+    /// <summary>
+    /// A Public Key Credential Source's generating authenticator determines at creation time whether the public key credential source is allowed to be backed up. 
+    /// Backup eligibility is signaled in authenticator data's flags along with the current backup state. 
+    /// Backup eligibility is a credential property and is permanent for a given public key credential source. 
+    /// A backup eligible public key credential source is referred to as a multi-device credential whereas one that is not backup eligible is referred to as a single-device credential.
+    /// <see cref="https://w3c.github.io/webauthn/#backup-eligibility"/>
+    /// </summary>
+    public bool BackupEligibility => _flags.HasFlag(AuthenticatorFlags.BE);
```
iamcarbon, line 51: Consider renaming IsBackupEligible

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -0,0 +1,86 @@
+﻿#nullable disable
+
+using System;
+using Fido2NetLib.Cbor;
+
+namespace Fido2NetLib.Objects
```
iamcarbon, line 6: Use file-scoped namespace.

Src/Fido2/Objects/AttestedCredentialData.cs
```diff
@@ -16,6 +16,7 @@ public sealed class AttestedCredentialData
     /// <see cref="https://www.w3.org/TR/webauthn/#attested-credential-data"/>
     /// </summary>
     private readonly int _minLength = Marshal.SizeOf(typeof(Guid)) + sizeof(ushort) + sizeof(byte) + sizeof(byte);
+    private readonly int _maxCredentialIdLength = 1023;
```
iamcarbon, line 19: change to const int?

Src/Fido2/Objects/AttestedCredentialData.cs
```diff
@@ -16,6 +16,7 @@ public sealed class AttestedCredentialData
     /// <see cref="https://www.w3.org/TR/webauthn/#attested-credential-data"/>
     /// </summary>
     private readonly int _minLength = Marshal.SizeOf(typeof(Guid)) + sizeof(ushort) + sizeof(byte) + sizeof(byte);
```
iamcarbon, line 18: change to const int _minLength = 12; // guid (8) + short (2) + byte (1) + byte (1) ?

Src/Fido2/DevelopmentInMemoryStore.cs
```diff
@@ -68,11 +68,52 @@ public Task<List<Fido2User>> GetUsersByCredentialIdAsync(byte[] credentialId, Ca
...
 public class StoredCredential
 {
+    /// <summary>
+    /// The type of the public key credential source.
+    /// </summary>
+    public PublicKeyCredentialType Type { get; set; } = PublicKeyCredentialType.PublicKey;
+    /// <summary>
+    /// The Credential ID of the public key credential source.
+    /// </summary>
+    public byte[] Id { get; set; }
```
iamcarbon, line 78: Add a return after each property for readability?

Src/Fido2/AuthenticatorAssertionResponse.cs
```diff
@@ -167,16 +182,210 @@ public static AuthenticatorAssertionResponse Parse(AuthenticatorAssertionRawResp
...
+                // https://w3c.github.io/webauthn/#sctn-device-publickey-attestation-calculations
+                try
+                {
+                    // This is a known device public key with a valid signature and valid attestation and thus a known device. Terminate these verification steps.
+                    _ = verifier.Verify(devicePublicKeyAuthenticatorOutput.AttStmt, devicePublicKeyAuthenticatorOutput.AuthData, devicePublicKeyAuthenticatorOutput.Hash);
+                    return devicePublicKeyAuthenticatorOutput.GetBytes();
+                }
+                catch
+                {
+                    // Some form of error has occurred. It is indeterminate whether this is a known device. Terminate these verification steps.
```
iamcarbon, line 383: Throw FidoVerificationException with inner exception?

Src/Fido2.Models/Objects/PublicKeyCredentialType.cs
```diff
@@ -11,5 +11,7 @@ namespace Fido2NetLib.Objects;
 public enum PublicKeyCredentialType
 {
     [EnumMember(Value = "public-key")]
-    PublicKey
+    PublicKey,
+    [EnumMember(Value = "invalid")]
```
iamcarbon, line 15: Add line break for readability.

Src/Fido2.Models/Objects/AssertionVerificationResult.cs
```diff
@@ -8,4 +8,17 @@ public class AssertionVerificationResult : Fido2ResponseBase
     public byte[] CredentialId { get; set; }
 
     public uint Counter { get; set; }
+
+    /// <summary>
+    /// The latest value of the signature counter in the authenticator data from any ceremony using the public key credential source.
+    /// </summary>
+    public uint SignCount { get; set; }
+    /// <summary>
```
iamcarbon, line 16: Add line break for readability.

iamcarbon:
Hi @aseigler -- anything left here?

Src/Fido2/Objects/AttestedCredentialData.cs
aseigler: Eh, I kind of like this one the way it is, matches the sizes mentioned in the spec.  Guid is 16 bytes, not 8.

Src/Fido2/AuthenticatorAssertionResponse.cs
aseigler: Yeah, I meant to go back and figure out what to throw for each of these but hadn't yet...and now I have to revisit again as some things have changed.

aseigler:
> Hi @aseigler -- anything functionality left here?

Not sure, need to do another pass on the DPK things to make sure I've got all the logic straight.  The passkey area I _think_ is good to go.

Src/Fido2/Objects/AttestedCredentialData.cs
iamcarbon: Ugh -- yes, Guid = 16. Mostly wanted to kill the call and make this a const, but given my own mistake above, maybe best as is! :)

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+
+    internal CborMap _attObj;
+
+    public byte[] AuthData => DataHelper.Concat(_dpkAuthDataPrefix, AaGuid.ToByteArray());
+
+    public byte[] Hash => DataHelper.Concat(DevicePublicKey.GetBytes(), Nonce);
+
+    public byte[] GetBytes() => _attObj.Encode();
+
+    public byte[] AuthenticationMatcher => DataHelper.Concat(AaGuid.ToByteArray(), DevicePublicKey.GetBytes());
```
iamcarbon, line 21: Do we need to consider the Endianness of the AaGuid when building the Authentication Matcher?

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
-        /// A CborMap encoded attestation statement.
-        /// </summary>
-        public CborMap AttStmt { get; }
-
-        /// <summary>
-        /// An optional boolean that indicates whether the attestation statement contains uniquely identifying information.
-        /// This can only be true when the `attestation` field of the extension input is "enterprise" and either the user-agent or the authenticator permits uniquely identifying attestation for the requested RP ID.
-        /// </summary>
-        public bool? EpAtt { get; private set; }
+        AaGuid = new Guid((byte[])attObjForDevicePublicKey["aaguid"]);
```
iamcarbon, line 25: Is aaguid BigEndian here?

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+        _attObj = attObjForDevicePublicKey;
     }
+
+    public DevicePublicKeyAuthenticatorOutput(byte[] attObjForDevicePublicKey) : this((CborMap)CborObject.Decode(attObjForDevicePublicKey)) { }
+
+    /// <summary>
+    /// The AAGUID of the authenticator. Can be used to identify the make and model of the authenticator.
+    /// <see cref="https://www.w3.org/TR/webauthn/#aaguid"/>
+    /// </summary>
+    public Guid AaGuid { get; private set; }
```
iamcarbon, line 43: Can we remove the private setter (making the backing fields readonly)

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+
+    /// <summary>
+    /// The credential public key encoded in COSE_Key format, as defined in 
+    /// Section 7 of RFC8152, using the CTAP2 canonical CBOR encoding form.
+    /// <see cref="https://www.w3.org/TR/webauthn/#credential-public-key"/>
+    /// </summary>
+    public CredentialPublicKey DevicePublicKey { get; private set; }
```
iamcarbon, line 50: Remove private setter

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+    /// 0x01 means "per-app" scope. Values other than 0x00 or 0x01 are reserved for future use.
+    /// </summary>
+    public uint Scope { get; private set; }
+
+    /// <summary>
+    /// An authenticator-generated random nonce for inclusion in the attestation signature.
+    /// If the authenticator chooses to not generate a nonce, it sets this to a zero-length byte string. 
+    /// See the note below about "randomNonce" for a discussion on the nonce's purpose.
+    /// </summary>
+    public byte[] Nonce { get; private set; }
```
iamcarbon, line 67: Consider backing with a ```private readonly byte[] _nonce```, and exposing over ```ReadOnlySpan<byte> Nonce => _nonce```; to prevent mutation.

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+
+    /// <summary>
+    /// Whether this key is scoped to the entire device, or a loosely-defined, narrower scope called "app". 
+    /// For example, a "device"-scoped key is expected to be the same between an app and a browser on the same device, while an "app"-scoped key would probably not be.
+    /// Whatever the scope, a device key is still specific to a given credential and does not provide any ability to link credentials.
+    /// Whether device-scoped or not, keys are still device-bound. I.e.an app-scoped key does not enjoy lesser protection from extraction.
+    /// A value of 0x00 means "entire device" ("all apps") scope. 
+    /// 0x01 means "per-app" scope. Values other than 0x00 or 0x01 are reserved for future use.
+    /// </summary>
+    public uint Scope { get; private set; }
```
iamcarbon, line 60: Remove private setter

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
+    /// <summary>
+    /// A CborMap encoded attestation statement.
+    /// </summary>
+    public CborMap AttStmt { get; }
+
+    /// <summary>
+    /// An optional boolean that indicates whether the attestation statement contains uniquely identifying information.
+    /// This can only be true when the `attestation` field of the extension input is "enterprise" and either the user-agent or the authenticator permits uniquely identifying attestation for the requested RP ID.
+    /// </summary>
+    public bool? EpAtt { get; private set; }
```
iamcarbon, line 84: Remove private setter.

Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs
```diff
@@ -1,86 +1,85 @@
...
 using System;
 using Fido2NetLib.Cbor;
 
-namespace Fido2NetLib.Objects
+
+public sealed class DevicePublicKeyAuthenticatorOutput
 {
-    public sealed class DevicePublicKeyAuthenticatorOutput
+    // https://w3c.github.io/webauthn/#sctn-device-publickey-attestation-calculations
+    internal readonly byte[] _dpkAuthDataPrefix = Convert.FromHexString("64657669636520626f756e64206b6579206174746573746174696f6e2073696700ffffffff");
```
iamcarbon, line 11: make internal static to avoid allocations per instance... or use compiler optimized ReadOnlySpan<byte> trick.

```sharp
internal static ReadOnlySpan<byte> _dpkAuthDataPrefix => new byte[] {
   100, 101, 118, 105, 99, 101, 32, 98, 111, 117, 110, 100, 32, 107, 101, 121, 32, 97, 116, 
   116, 101, 115, 116, 97, 116, 105, 111, 110, 32, 115, 105, 103, 0, 255, 255, 255, 255
 };
 ```

**Review by iamcarbon — CHANGES_REQUESTED** [review]

File: Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs (hunk 12 above)
Comment (aseigler) [reply Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs:25]: Probably.  I'll add the same logic as we use elsewhere.

File: Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs (hunk 11 above)
Comment (aseigler) [reply Src/Fido2/Objects/DevicePublicKeyAuthenticatorOutput.cs:21]: Not as long as we're consistent with the storage during registration vs logon, which we will be.

File: Src/Fido2/Objects/AttestedCredentialData.cs (hunk 6 above)
Comment (aseigler) [reply Src/Fido2/Objects/AttestedCredentialData.cs:16]: I am just going to go with a `readonly int` value of 20.  This isn't spec level check, it's a sanity check added so we can tell when the conformance test that sends a value that is way too small we can tell that is specifically the test we passed and not some other problem.

**Comment by aseigler** [comment]
Still need to make RP configurable BE/BS options and plumb those into registration and login flows before passkey is done.

DPK looks like it has object names that have been updated. 

Reminder to create CredentialRecord object, move existing StoredCredential members there and hand around that credential record instead of parts of it in and out of login flow.

Reminder to remove the rest of the references to token binding.

File: Demo/Startup.cs [hunk 19]
```diff
@@ -52,6 +52,8 @@ public void ConfigureServices(IServiceCollection services)
             options.Origins = Configuration.GetSection("fido2:origins").Get<HashSet<string>>();
             options.TimestampDriftTolerance = Configuration.GetValue<int>("fido2:timestampDriftTolerance");
             options.MDSCacheDirPath = Configuration["fido2:MDSCacheDirPath"];
+            options.AllowBackupEligibleCredential = Configuration.GetValue<bool>("fido2:allowBackupEligibleCredential");
```
rudiv, line 55: I would argue that these be allowed by default (ie. `Configuration.GetValue<bool>("fido2:allowBackupEligibleCredential", true);`) to support a happier migration path. Some MPs may not care about these statuses at all, and so their existing configuration would break?

aseigler: > I would argue that these be allowed by default to support a happier upgrade path. Some RPs may not care about these statuses at all, and so their existing configuration would break?
> 
> I know this comment is against the Demo project, but the [Configuration default](https://github.com/aseigler/fido2-net-lib/blob/feat-passkey/Src/Fido2.Models/Fido2Configuration.cs#L120) would also be false if this wasn't decided / set at startup.

I hear ya.  This is a enterprise vs consumer setting in my eyes, and I personally am enterprise focused.  When this PR is in a release it will be part of a major update, and we are expecting that will contain some breaking changes.  Regardless of which way it goes by default, there will be notes to help guide implementers.

rudiv: I agree, I think my main concern is that this may block typical use cases.

For an example (even thinking enterprise), you may wish for your users to have a passkey that's backed up and set this setting. However, they would then be unable to add a physical key (Yubikey et al). I know this is the whole point of "Passkeys" to not need it, but I feel that it's more of a per-key decision than one that should be set globally for the RP?

abergs: @aseigler I agree that we should default to true. Given that the average developer might not know what this option do (and should be True for them), while I think we can expect enterprise developers have more knowledge and know to turn it off because of their own research/policy for deploying fido2.

iamcarbon:
@aseigler Anything else left here before we merge?

aseigler:
> @aseigler Anything else left here before we merge?

I think the only thing that hasn't been gotten to in the list was moving StoredCredential to CredentialRecord but that can be done separately, no need to hold this up for that.

Demo/Startup.cs
aseigler: > 
They would actually be able to register a typical security key with this setting true or false, as those do not set BE flag. Currently only certain platform authenticators do. It's undefined how the platform is protecting those BE keys and where they are going to be restored to. That's why in a situation where the RP cares (enterprise), they might wish to restrict BE keys from being registered at all on the platform, or maybe allow some blessed authenticators but not others (would require additional work to support this). I guess there might be a reason an RP would only want to allow BE keys? That's not the result or intent of this implementation, at least currently.

rudiv: Understand the Enterprise point of view.

From a more consumer-focused point of view (ie. specifically the well marketed "Passkeys" aspect) we may want to accept _only_ Passkeys at first, to ensure that users are not able to lose access very easily. Whilst it's not perfect and most authenticators don't give on where the keys are from / stored, knowing that they're backed up allows it to be reasonably assumed that the key is durable and isn't going to disappear like in the case of a physical key loss. For example, we explicitly advise the user to register with a backed up authenticator (specifically iCloud backed Passkeys or their Android phone).

I might be wrong in assuming that this is where this library is going to see most of its use in the short-mid term, but that's my 2c on why this default would be the wrong one.

Dannzak:
Hey! Any news on this?

Src/Fido2/AuthenticatorAttestationResponse.cs
```diff
@@ -157,6 +169,119 @@ public static AuthenticatorAttestationResponse Parse(AuthenticatorAttestationRaw
...
+        //     the Relying Party SHOULD fail the registration ceremony.
+        //     This implementation throws if the outputs are not trustworthy for a particular attestation type.
+
+        return new AttestationVerificationSuccess
+        {
+            Type = Raw.Type,
+            Id = authData.AttestedCredentialData.CredentialID,
+            PublicKey = authData.AttestedCredentialData.CredentialPublicKey.GetBytes(),
+            SignCount = authData.SignCount,
+            //Transports = result of response.getTransports();
```
abergs, line 213: Why do we not return Transports anymore?

Src/Fido2/DevelopmentInMemoryStore.cs
```diff
@@ -68,12 +68,70 @@ public Task<List<Fido2User>> GetUsersByCredentialIdAsync(byte[] credentialId, Ca
...
+    public byte[] AttestationClientDataJSON { get; set; }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public List<byte[]> DevicePublicKeys { get; set; }
+
     public byte[] UserId { get; set; }
+
     public PublicKeyCredentialDescriptor Descriptor { get; set; }
```
abergs, line 125: Should we depricate the descriptor or do we still have a use for it? (Now that we've added Id, PublicKey, Type to the class itself)

Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
```diff
@@ -9,18 +9,40 @@ namespace Fido2NetLib.Objects;
...
-
-    public X509Certificate2[] AttestationCertificateChain { get; set; }
+    /// <summary>
+    /// The type of the public key credential source.
+    /// </summary>
+    public PublicKeyCredentialType Type { get; set; } = PublicKeyCredentialType.PublicKey;
+    /// <summary>
+    /// The Credential ID of the public key credential source.
+    /// </summary>
+    public byte[] Id { get; set; }
```
abergs, line 24: Do we want to use the same JsonConverter as we do on PublicKey?

Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
```diff
@@ -9,18 +9,40 @@ namespace Fido2NetLib.Objects;
...
+    public AuthenticatorTransport[] Transports { get; set; }
+    /// <summary>
+    /// The value of the BE flag when the public key credential source was created.
+    /// </summary>
+    public bool BE { get; set; }
+    /// <summary>
+    /// The value of the attestationObject attribute when the public key credential source was registered. 
+    /// Storing this enables the Relying Party to reference the credential's attestation statement at a later time.
+    /// </summary>
+    public byte[] AttestationObject { get; set; }
```
abergs, line 42: Do we want to use JsonConverter on this and `AttestationClientDataJSON`?

abergs (approved):
Great work on these two large features. I have some comments, but I'm happy with the changes.

Src/Fido2/AuthenticatorAttestationResponse.cs
aseigler: We didn't return transports before, in fact we didn't even collect it in the JavaScript at all, I added in [this](https://github.com/passwordless-lib/fido2-net-lib/blob/53caf81da894691ad10eefcce5fd5979b4a8e6ed/Demo/wwwroot/js/custom.register.js#L129) but it looks like it needs to be added to all of the other registration JavaScript files, and collect transports in [AuthenticatorAttestationRawResponse.ResponseData](https://github.com/passwordless-lib/fido2-net-lib/blob/53caf81da894691ad10eefcce5fd5979b4a8e6ed/Src/Fido2.Models/AuthenticatorAttestationRawResponse.cs#L26), and then populate transports in the [controller from the raw response](https://github.com/passwordless-lib/fido2-net-lib/blob/53caf81da894691ad10eefcce5fd5979b4a8e6ed/Demo/Controller.cs#L131).  There are a few reasons to collect transports, to help with user facing dialogs, but also to compare what transports are being used against a policy, or comparing what transports should be available according to the authenticator metadata.  Basically this is incomplete at the moment.

Demo/TestController.cs
```diff
@@ -154,30 +154,33 @@ public IActionResult AssertionOptionsTest([FromBody] TEST_AssertionClientParams
...
         // 5. Make the assertion
-        var res = await _fido2.MakeAssertionAsync(clientResponse, options, creds.PublicKey, storedCounter, callback, cancellationToken: cancellationToken);
+        var res = await _fido2.MakeAssertionAsync(clientResponse, options, creds.PublicKey, creds.DevicePublicKeys, storedCounter, callback, cancellationToken: cancellationToken);
 
         // 6. Store the updated counter
-        DemoStorage.UpdateCounter(res.CredentialId, res.Counter);
+        _demoStorage.UpdateCounter(res.CredentialId, res.Counter);
+
+        if (res.DevicePublicKey is not null)
+            creds.DevicePublicKeys.Add(res.DevicePublicKey);
```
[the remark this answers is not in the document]
abergs: Hmm... I think I understand @aseigler. Am I right to think that before saving the DPK the RP would perhaps check some policy or trigger a risk based ceremony before trusting that DPK and then saving/updating the storage with `res.DevicePublicKey` but in our demo we just add it directly to your in memory storage?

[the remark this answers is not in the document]
aseigler: Yeah, basically same as registering a new credential against an existing user, but _way_ more complicated.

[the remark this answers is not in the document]
abergs: Alright makes sense, I was just confused about how we used it in our demo. When I get the time I'll add a `// TODO: Check policy if we allow new devices, if-so, store it to db`

Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
aseigler: Already done in AuthenticatorAttestationRawResponse, no?

Src/Fido2.Models/Objects/AttestationVerificationSuccess.cs
aseigler: Already done in AuthenticatorAttestationRawResponse, no?

PR #363, opened 2023-01-16
Fix Conformance Test

iamcarbon:
They certinfa url was updated. This PR updates the url and improves the error message if it goes down in the future. This PR also updates the test dependencies to their latest versions.

@abergs @aseigler Ready for review!

Also, congratulations on the BitWarden acquisition!


abergs:
This is the first congratulation I had that is accompanied with a Pull Request. I consider it the best kind @iamcarbon

abergs:
LGTM :shipit:

PR #372, opened 2023-01-28
RFC : Make models immutable

iamcarbon:
The purpose of this PR is to start a discussion on whether we want to introduce immutable models into the library for the 4.0 release. If we decide to move forward on this, this would be done after the Passkeys merge to prevent conflicts. 

Immutable models ensure:

- instances are validated at construction time according to the specification
- instances aren't modified and remain valid for their entire lifetime
- the instances can be safely reused

Apple & Android also have exposed their Fido2 implementation over immutable models. 

for example:

https://developers.google.com/android/reference/com/google/android/gms/fido/fido2/api/common/PublicKeyCredentialUserEntity#getIcon()

https://developer.apple.com/documentation/authenticationservices/asauthorizationsecuritykeypublickeycredentialdescriptor


iamcarbon:
@abergs @aseigler Let me know your thoughts when you get a moment to look over!

PR #374, opened 2023-02-04
Make options de-/serializable by dotnet

### LINKED ISSUE #371: Deserializing in frontend: ctor gets trimmed
So I'm basically far enough to actually deploy my Blazor WASM sample to the internet, except in the published version the deserializing throws a NotSupportedException.

Turn out publishing to WASM trims the constructor or Metadata about the constructor about the `CredentialCreationOptions` away, leading to:

`System.NotSupportedException: DeserializeNoConstructor, JsonConstructorAttribute, Fido2NetLib.CredentialCreateOptions `

Added an explicit default ctor to CredentialCreationOptions, added `[JsonConstructor]` to it, built a local package, and published with that - still that error.

Removed \<IsTrimmable> from Fido2.Models **and it worked**.

So how can we make dotnet keep the ctor but still make the assembly trimmable? Especially in the Blazor frontend that needs longer to load than other clients it would be nice to have trimming. I would've thought that both the `FidoModelSerializerContext` and the `JsonConstructor` each should have solved this problem by themselves already.

Regenhardt:
closes #371

### DIFF
--- a/Src/Fido2.Models/CredentialCreateOptions.cs
+++ b/Src/Fido2.Models/CredentialCreateOptions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
@@ -118,6 +119,7 @@ public sealed class PubKeyCredParam
     /// Constructs a PubKeyCredParam instance
     /// </summary>
     [JsonConstructor]
+    [DynamicDependency(DynamicallyAccessedMemberTypes.PublicConstructors, typeof(PubKeyCredParam))]
     public PubKeyCredParam(COSE.Algorithm alg, PublicKeyCredentialType type = PublicKeyCredentialType.PublicKey)
     {
         Type = type;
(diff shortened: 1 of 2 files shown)


Regenhardt:
Turns out PubKeyCredParams are immutable so giving the serializer a default constructor stops the deserialization form setting the properties. Guess I have to find out more about the `ConstructorContainsNullParameterNames` thing.

Regenhardt:
Ok this one works correctly now, telling the linker that if we keep the ctor we wanna keep the whole ctor. Will definitely open an issue with Microsoft to check if this is planned behaviour or if the linker needs a fix.

Regenhardt:
Or should we rather add these changes to a new project, which would then also contain [this](https://gitlab.com/Regenhardt/fido2-webauthn-blazor/-/blob/master/WebAuthn/WebAuthn.cs) js wrapper and [this](https://gitlab.com/Regenhardt/fido2-webauthn-blazor/-/blob/master/Client/wwwroot/webauthn.ts) file, even though a bit cleaned up and with the nuget package actually adding the transpiled js file?

Regenhardt:
Something like Fido2.Blazor or Fido2.WebAssembly or Fido2.BlazorWebAssembly like the sdk.

abergs:
Yes, moving things into a Blazor project (and small demo) would be ideal for changes like these.

Regenhardt:
Since this became bigger as expected, I made a new branch: #379

PR #376, opened 2023-02-06
Hackaton updates

sebfort:


sebfort:
These are just tests and we do NOT want to merge them into original repository.

Regenhardt:
Either mark as draft then or, which would be the better solution, create the PR into sebfort:master instead?

abergs:
Hi @sebfort, I appreciate your hackathon and experimentation but I think you've misstakenly opened a PR against our repo instead of your own. Will Close this PR as I don't think you want us to review your changes and include them in the repo.

PR #383, opened 2023-03-07
Modify the csproj file so that it can be deployed correctly to an Azure app service

jameswestgate:
When trying to deploy the demo to an Azure app service, you encounter the following issue:

https://github.com/dotnet/sdk/issues/251 
https://stackoverflow.com/questions/43320772/the-publish-target-is-not-supported-without-specifying-a-target-framework

The csproj file needs to be edited so that the instead of <TargetFrameworks> you use <TargetFramework>. Since only .net 6.0 is targeted, this isn't a problem.


Regenhardt:
I have temporarily deployed a demo app using these packages without having to modify their project files, it seems to work fine for me:

https://fido2-blazordemo.azurewebsites.net

jameswestgate:
> 

Extra commits have been added to the pull request. Apologies. Im also guessing that perhaps this issue is being resolved by an update to VS. The only relevant commit is https://github.com/passwordless-lib/fido2-net-lib/pull/383/commits/66566c82489f7700235073c6afca237514c7a449

I suggest we close this pull request.

PR #393, opened 2023-05-17
Make Origins readonly

### LINKED ISSUE #392: adding a origin to Origins directly should throw a error
It should be made clear that this code doesn't work. With an error/warning message when setting options.Origins to an empty set. Or the error: "Fully qualified origin {} of {} not equal to fully qualified original origin {,} ({})" could be elaborated to remind the user of checking how the HashSet had been inited 
```
Services.AddFido2(options=>{
        //...
        options.Origins = new HashSet<string>();
        options.Origins.Add("http://localhost:5091");
       //...
})
```

Regenhardt:
Closes #392

### DIFF
--- a/Src/Fido2.Models/Fido2Configuration.cs
+++ b/Src/Fido2.Models/Fido2Configuration.cs
@@ -7,8 +7,8 @@ namespace Fido2NetLib;
 
 public class Fido2Configuration
 {
-    private ISet<string> _origins;
-    private ISet<string> _fullyQualifiedOrigins;
+    private IReadOnlySet<string> _origins;
+    private IReadOnlySet<string> _fullyQualifiedOrigins;
 
     /// <summary>
     /// Create the configuration for Fido2.
@@ -34,7 +34,7 @@ public class Fido2Configuration
     public int ChallengeSize { get; set; } = 16;
 
     /// <summary>
-    /// The effetive domain of the RP. Should be unique and will be used as the identity for the RP.
+    /// The effective domain of the RP. Should be unique and will be used as the identity for the RP.
     /// </summary>
     public string ServerDomain { get; set; }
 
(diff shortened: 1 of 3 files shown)


Src/Fido2.Models/Fido2Configuration.cs
```diff
@@ -56,20 +56,15 @@ public Fido2Configuration()
...
-                _origins = new HashSet<string>();
-
-                // Since we're depricating Origin we ease the transition to move the value automatically, unless its null
+                // Since we're deprecating Origin we ease the transition to move the value automatically, unless its null
 #pragma warning disable CS0618 // Type or member is obsolete
-                if (Origin != null)
-                {
-                    _origins.Add(Origin);
-                }
+                _origins = new HashSet<string>(new[]{Origin});
```
abergs, line 67: This would add null to `Origins` if `Origin` is null, is that really what we want? I would prefer it to be empty.

Regenhardt: Oh yes definitely, I updated it.

Src/Fido2.Models/Fido2Configuration.cs
```diff
@@ -56,20 +56,17 @@ public Fido2Configuration()
...
-                _origins = new HashSet<string>();
-
-                // Since we're depricating Origin we ease the transition to move the value automatically, unless its null
+                // Since we're deprecating Origin we ease the transition to move the value automatically, unless its null
 #pragma warning disable CS0618 // Type or member is obsolete
-                if (Origin != null)
-                {
-                    _origins.Add(Origin);
-                }
+                _origins = Origins != null
```
abergs, line 67: Looks good, but small misstake null checking `Origins` instead of `Origin`

Regenhardt: Guess I shouldn't work on official projects this late. Thanks for keeping an eye out.

aseigler:
No objections, makes sense to me.

Regenhardt:
Like this? Although FullyQualifiedOrigins might aswell be a { get; private set; } property now. Should I additionally document the breaking change somewhere?

abergs:
@Regenhardt It will be documented automatically in the Release notes.

I will do a final review and merge as soon as I have a moment. Traveling this week.

abergs:
@Regenhardt This now has a bit of conflicts because we brought some other changes in. Would you mind rebasing and sorting it out and we're pretty good to go?

Regenhardt:
Rebased it

abergs:
Re-ran all tests, seems good now.

abergs (approved):

PR #411, opened 2023-07-28
Add CredentialPropertiesOuput (RK Support)

abergs:
Browsers respond with a a object.

- Add CredentialPropertiesOutput
- Use .ID instead of null CredentialId


abergs:
Yeah, it's in the works.

PR #414, opened 2023-07-31
Enable C# 11 and embed base64 encoded certificates as UTF-8 data

iamcarbon:
Is a bit faster to parse UTF-8 data directly than convert from UTF-16. 

NOTE, this uses the built-in support for static data initialization. Accessing ROOT_CERT is non-allocating and refers directly to the data.

This PR also simplifies a check using C# 11 list patterns.


iamcarbon:
Ahh, CI doesn't like C# 11. I'll look into this a bit later.

iamcarbon:
future note: We should also consider cross-targeting and testing against net8.0 after the next preview. This will also provide us with the latest System.Text.Json source generation improvements.

iamcarbon:
@abergs Ready for review.

abergs (approved):
Looks good, thank you @iamcarbon

PR #422, opened 2023-08-21
[WIP] Improve Code Quality

iamcarbon:
- Remove platform specific logic that is no longer necessary on .NET core
- Add implicit casting to CborByteString and CborTextString to their respective types (byte[] and string)
- Use optimized OperatingSystem classes for platform specific logic
- Minor test cleanup
- Improve AndroidKey error messages
- Improve CredentialPublicKey test coverage


iamcarbon:
@abergs Ready for review

Src/Fido2.Models/COSETypes.cs
```diff
@@ -186,7 +186,7 @@ public enum EllipticCurve
         /// </summary>
         Ed448 = 7,
         /// <summary> 
-        /// secp256k1 (pending IANA - requested assignment 8)
+        /// secp256k1
```
iamcarbon, line 189: This has an assignment now.

Src/Fido2/AttestationFormat/AndroidSafetyNet.cs
```diff
@@ -29,26 +29,31 @@ public override (AttestationType, X509Certificate2[]) Verify(VerifyAttestationRe
...
-        if (!(request.AttStmt["response"] is CborByteString { Length: > 0 }))
-            throw new Fido2VerificationException("Invalid response in SafetyNet data");
+        if (!(request.AttStmt["response"] is CborByteString { Length: > 0 } responseByteString))
+            throw new Fido2VerificationException(Fido2ErrorMessages.InvalidSafetyNetResponse);
 
-        var response = (byte[])request.AttStmt["response"]!;
-        var responseJWT = Encoding.UTF8.GetString(response);
+        var responseJwt = Encoding.UTF8.GetString(responseByteString);
 
-        if (string.IsNullOrWhiteSpace(responseJWT))
```
iamcarbon, line 41: The whitespace case now returns a generic malformed JWT error message.

Src/Fido2/AttestationFormat/AndroidSafetyNet.cs
```diff
@@ -29,26 +29,31 @@ public override (AttestationType, X509Certificate2[]) Verify(VerifyAttestationRe
...
+        byte[] jwtHeaderBytes;
 
-        string jwtHeaderString = jwtParts[0];
+        try
+        {
+            jwtHeaderBytes = Base64Url.Decode(jwtComponents[0]);
+        }
+        catch (FormatException)
+        {
+            throw new Fido2VerificationException(Fido2ErrorMessages.MalformedSafetyNetJwt);
```
iamcarbon, line 53: We also returned a malformed JWT error message when the header is invalid

Src/Fido2/AttestationFormat/FidoU2f.cs
```diff
@@ -36,15 +36,9 @@ public override (AttestationType, X509Certificate2[]) Verify(VerifyAttestationRe
         var pubKey = attCert.GetECDsaPublicKey()!;
         var keyParams = pubKey.ExportParameters(false);
 
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
abergs, line 39: Can't remember the case we used this for, but I guess we will notice if something breaks.

Src/Fido2/Extensions/CryptoUtils.cs
```diff
@@ -139,7 +139,7 @@ public static byte[] SigFromEcDsaSig(byte[] ecDsaSig, int keySize)
 
         // Create buffer to copy R into 
         Span<byte> p1363R = coefficientSize <= 64
-            ? stackalloc byte[coefficientSize]
+            ? (stackalloc byte[64]).Slice(0, coefficientSize)
```
abergs, line 142: What's happening here? Why are we always allocating 64 bytes now?

Test/Extensions/CertInfoHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+        ReadOnlySpan<byte> extraData,
+        ReadOnlySpan<byte> clock,
+        ReadOnlySpan<byte> resetCount,
+        ReadOnlySpan<byte> restartCount,
+        ReadOnlySpan<byte> safe,
+        ReadOnlySpan<byte> firmwareRevision,
+        ReadOnlySpan<byte> tPM2BName,
+        ReadOnlySpan<byte> attestedQualifiedNameBuffer)
+    {
+        var stream = new MemoryStream();
```
abergs, line 18: Do we need a using here?

Test/Extensions/PubAreaHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+        ReadOnlySpan<byte> policy,
+        ReadOnlySpan<byte> symmetric,
+        ReadOnlySpan<byte> scheme,
+        ReadOnlySpan<byte> keyBits,
+        ReadOnlySpan<byte> exponent,
+        ReadOnlySpan<byte> curveID,
+        ReadOnlySpan<byte> kdf,
+        ReadOnlySpan<byte> unique = default)
+    {
+        var raw = new MemoryStream();
```
abergs, line 24: Do we need a using here?

Src/Fido2/Extensions/CryptoUtils.cs
iamcarbon: It can be slightly faster to allocate small regions from the stack using a const value and slicing it to size than to use a variable. 

There's a bit of discussion on this here: https://github.com/dotnet/docs/issues/28823 and a general recommendation here: https://vcsjones.dev/stackalloc/.

However, it looks like there's some debate on the thresholds where this is an improvement, and whether the JIT can be improved in the future to make this irrelevant.

https://github.com/dotnet/runtime/issues/89478

Given that our preceding <= check makes the safety concern irrelevant, the performance impact is 1-2ns, and that this might be optimized in the future -- will revert this for clarity.

Test/Extensions/CertInfoHelper.cs
iamcarbon: Technically, MemoryStreams don't need disposed -- as they're backed by managed memory that can be safely garbage collected. However, I like adding using statements to them anyway to match other stream types. Will update.

Here's the official documentation here:

```
This type implements the [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-7.0) interface, but does not actually have any resources to dispose. This means that disposing it by directly calling [Dispose()](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-7.0#system-idisposable-dispose) or by using a language construct such as using (in C#) or Using (in Visual Basic) is not necessary.
```

https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=net-7.0

abergs: @iamcarbon Oh, TIL. I was taught to always dispose streams and never try to be clever, which was a good lesson but overly cautious.

abergs:
@iamcarbon I understood your making some small changes? I'm OK either way.

iamcarbon:
@abergs All set. I know these small changes can get tedious to review -- and I appreciate you for for taking the time and for the feedback.

The 4.0 release is shaping up nicely!

Src/Fido2/Objects/CredentialPublicKey.cs
```diff
@@ -93,7 +93,7 @@ public bool Verify(ReadOnlySpan<byte> data, ReadOnlySpan<byte> signature)
         throw new InvalidOperationException($"Missing or unknown kty {_type}");
     }
 
-    internal RSA CreateRsa()
+    internal RSA CreateRSA()
```
iamcarbon, line 96: This is a specific case for breaking the .NET naming rules, since the RSA class name is also capitalized.

abergs (approved):

abergs:
@iamcarbon Your contributions are highly valued!

aseigler:
> @iamcarbon Your contributions are highly valued!

^ Extremely!

PR #425, opened 2023-08-24
dotnet format

abergs:
- Added pragmas
- dotnet format
- Disabled var_elsewhere

With this I think we can enforce dotnet format on all changes via `dotnet format --verify-no-changes`


abergs:
@iamcarbon I think this works OK. Any thoughts?

abergs:
@iamcarbon or @aseigler Would be interesting to see if you clone this branch and run 'dotnet format', do you see no changes? 

The build was triggering changes on windows but not mac.

Src/Fido2.Ctap2/Commands/AuthenticatorClientPinCommand.cs
```diff
@@ -96,6 +96,7 @@ public sealed class AuthenticatorClientPinCommand : CtapCommand
 
 public enum AuthenticatorClientPinSubCommand
 {
+    #pragma warning disable format
```
iamcarbon, line 99: Oh, this is nice! 👍

Issue #426, opened 2023-08-30
Clean up and break things


Regenhardt:
As previously mentioned, with breaking changes coming anyway, we might as well clean up some stuff. So here's things I understood from [over here](https://github.com/passwordless-lib/fido2-net-lib/pull/379#issuecomment-1640340625) we could change, I'll try to make different points and work within those so I don't get confused too much.

# 1. StoredCredential.CredType

> `StoredCredential.Type` is a `PublicKeyCredentialType`, which currently can only be `public-key`. The `StoredCredential.CredType` is meant to store the attestation `fmt` string for...reasons I don't remember. Probably a bad name, this is a good time to rename it.

This directly maps to `AttestationVerificationSuccess.CredType`. If I understand this correctly, it maps to the spec's `attestationFormat`(see https://w3c.github.io/webauthn/#sctn-generating-an-attestation-object), so how about I rename `CredType` to `AttestationFormat` plus documentation like "Format of the attestation statement, e.g. fido-u2f or android-key" with that link?

<hr />

## Update 2024-01-12

Done in #427

# 2. Id/CredentialId

>    What's the difference between `AttestationVerificationSuccess.Id` and `AssertionVerificationResult.CredentialId`, which `AttestationVerificationSuccess` inherits? Seem redundant again.

> I think those are redundant. A few times I tried to move everything to use the `AssertionVerificationResult.CredentialId` and remove `AttestationVerificationSuccess.Id` but I kept getting wrapped around the axles. This is a good time to fix that as well, if we can figure out how to do it cleanly.

* These seem to be used internally only: 
  * `.Id` is created on attestation, then stored with the new credential, both as `.Id` and `.Descriptor.Id` -> redundant again?
  * `.CredentialId` is created on successful assertion, used to up the counter, literally used a index to find the stored `.Descriptor.Id`, TestController also stores it in the `.Descriptor.Id`, I'm confident it's the same as `.Id`.

Was speaks against shoving everything into `.CredentialId`? Tbh. I'm having a hard time finding this one in the spec, it's possible there was something along the lines of `public byte[] Id => this.CredentialId;` in there basically but in natural language.  
If you think this would ok spec-wise, I'll try to make everything use `CredentialId`.

<hr />

## Update 2024-01-12

`AttestedCredentialData` (lib-input in Attestation/Assertion authenticator responses) has `CredentialId`  
`RegisteredPublicKeyCredential` (lib-output when cred was created / after attestation) has `Id`  
[...]

PR #427, opened 2023-08-30
Breakout development services & improve naming

iamcarbon:
- Moves the DevelopmentInMemoryStore, StoredCredential, and ConformanceMetadataService into a new Fido2.Development project.

- Renames a few properties on StoredCredential for clarity.
- Removes duplicate properties on StoredCredential (preferring specification name)


BlazorWasmDemo/Client/BlazorWasmDemo.Client.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <LangVersion>11</LangVersion>
```
Regenhardt, line 7: Better remove this property from all the projects and update the one in the global Directory.Build.props instead.

Src/Fido2.Development/StoredCredential.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public List<byte[]> DevicePublicKeys { get; set; }
+
+    public byte[] UserId { get; set; }
+
+    public PublicKeyCredentialDescriptor Descriptor { get; set; }
+
+    public byte[] UserHandle { get; set; }
+
+    public string CredType { get; set; }
```
Regenhardt, line 59: Could rename this to `AttestationFormat`, see #426.

Regenhardt:
This will definitely complete 3) from #426, nice.

Src/Fido2.Development/StoredCredential.cs
iamcarbon: Agreed that AttestationFormat is a better name here. Updated.

iamcarbon:
@abergs Ready for review. This partially addresses the feedback received in #427

abergs:
Thanks @iamcarbon and @Regenhardt - I've done a first pass, but will need more time to review this as it's extensive changes. Combined with a busy schedule, it make take some extra time before I'm able to merge. Thanks for the contributions and patience.

abergs:
Just an update, I will start work on this.

jonashendrickx (approved):

iamcarbon:
@abergs I think we have two options on CredentialId. 

1. Drop it from AttestationVerificationSuccess (it was never set)
2. Rename Id to CredentialId. 

My thinking on keeping Id (and dropping CredentialId) is explained in #428. The meaning of Id could also be clarified renaming the type to something ending in Credential.

BlazorWasmDemo/Client/BlazorWasmDemo.Client.csproj
abergs: @iamcarbon Agreed, but could be separate PR.

Src/Fido2/Objects/AuthenticatorData.cs
```diff
@@ -79,7 +79,7 @@ public AuthenticatorData(byte[] rpIdHash, AuthenticatorFlags flags, uint signCou
     /// Backup state is signaled in authenticator data's flags and can change over time.
     /// <see cref="https://w3c.github.io/webauthn/#backup-state"/>
     /// </summary>
-    public bool BackupState => _flags.HasFlag(AuthenticatorFlags.BS);
+    public bool IsBackedUp => _flags.HasFlag(AuthenticatorFlags.BS);
```
abergs, line 82: I was contemplating if we should always keep using Spec terms on places you added `IsBackedUp`.

I decided the readability is worth not aligning to spec (as long as we mention BS/BE in XML).

abergs (approved):

abergs:
@iamcarbon While this PR is accepted as is - It would be nice to change the name of `AttestationVerificationSuccess` to `RegisteredPublicKeyCredential ` - it would make DX a bit better I think.

iamcarbon:
@abergs All set.

PR #428, opened 2023-09-28
CredentialId is no longer mapped

jonashendrickx:
Are we still using CredentialId? It is no longer mapped, and therefore null. Might have been a breaking change from @aseigler ?

https://github.com/passwordless-lib/fido2-net-lib/commit/7b92f4a019a9ed24793eedb8016408b44e9ee858#diff-1aab08f33f616a78e8c75fa260f05ac7726414bab4422b432c6c21c0d661dee4

### DIFF
--- a/Src/Fido2/AuthenticatorAttestationResponse.cs
+++ b/Src/Fido2/AuthenticatorAttestationResponse.cs
@@ -188,6 +188,7 @@ public sealed class AuthenticatorAttestationResponse : AuthenticatorResponse
         {
             Type = Raw.Type,
             Id = authData.AttestedCredentialData.CredentialID,
+            CredentialId = authData.AttestedCredentialData.CredentialID,
             PublicKey = authData.AttestedCredentialData.CredentialPublicKey.GetBytes(),
             SignCount = authData.SignCount,
             //Transports = result of response.getTransports();


jonashendrickx:
@dotnet-policy-service agree

Regenhardt:
Would be nice to have everything use `.CredentialId`, would you try and remove the `.Id` property used in some places in favour of `.CredentialId` of the base type?  
If not I'm ok with that too, in that case this would just be the first step in converging everything to `.CredentialId`.

abergs:
@Regenhardt Since I'm embarking on the review of the other #427 - is this not changed in that PR?

jonashendrickx:
> @Regenhardt Since I'm embarking on the review of the other #427 - is this not changed in that PR? I imagined it would be.

It is removed in #427 completely. If #427 is approved, we can abandon this PR.

abergs:
@jonashendrickx Yeah I figured.

Regenhardt:
As far as I can see, #427 renames `CredentialID` to `CredentialId` - the d is small now. 

I hoped this might be an attempt at solving 2) from #426, removing the duplicate `AttestationVerificationSuccess.Id` in favour of the `AttestationVerificationResult.CredentialId` of its base type.

iamcarbon:
@Regenhardt #427 Removed the base type from AttestationVerificationSuccess to eliminate the unused CredentialId.

Since the result of AuthenticatorAttestationResponse.VerifyAsync is a newly registered PublicKeyCredential -- keeping the Id property still make sense. We might consider renaming AttestationVerificationSuccess to RegisteredPublicKeyCredential (which infers verification was successful), and also make the Id's meaning clear.

When verifying or making an assertion, we're verifying an existing credential that we reference. The credential is referenced by VerifyAssertionResult -- so the CredentialId property makes sense here.

It we keep the class name (and don't consider this result as an extended PublicKeyCredential), renaming to CredentialId would also be reasonable. 

@abergs ^ my thinking on keeping the Id on AttestationVerificationSuccess. Let me know your thoughts, and whether we should also consider renaming this type to RegisteredPublicKeyCredential.

abergs:
@iamcarbon I agree with your analysis and I asked for changing the name of `AttestationVerificationSuccess` to `RegisteredPublicKeyCredential ` in #427.

However - I do think @Regenhardt has a point that we should ponder going forward, that is: Calling it `CredentialId` throughout is indeed helpful - but it's not clear cut as it gets complicated when we're trying to stay as close to the spec as possible. 

Closing this as it is adressed in #427.

PR #433, opened 2023-10-09
Rename CredentialMakeResult to MakeNewCredentialResult and adjust its namespace 

### LINKED ISSUE #432: Namespace inconsistency between CredentialMakeResult and AssertionVerificationResult
`CredentialMakeResult` should not be a class child of `Fido2NetLib.Fido2`. `AssertionVerificationResult` is at the root of the namespace `Fido2NetLib.Objects`. We should probably move `CredentialMakeResult` under the namespace `Fido2NetLib.Objects`.

joegoldman2:
Fixes #432

### DIFF
--- a/Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
+++ b/Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
@@ -1,11 +1,5 @@
 ﻿#nullable enable
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
 namespace Fido2NetLib.Objects;
 
 /// <summary>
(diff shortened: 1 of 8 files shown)


iamcarbon:
@joegoldman2 Let's rename this type to MakeNewCredentialResult while we are here.

iamcarbon:
👍

Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
```diff
@@ -0,0 +1,26 @@
+﻿#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
aseigler, line 5: Some of these usings seem unused?

aseigler (changes requested):

Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
joegoldman2: Good catch! In fact, they are all unused. I remove them.

abergs (approved):
LGTM. @aseigler anything or are we good to merge?

aseigler:
> LGTM. @aseigler anything or are we good to merge?

I'm happy if you're happy!

PR #434, opened 2023-10-09
Use DateTimeOffset

### LINKED ISSUE #274: Use DateTimeOffset
https://github.com/passwordless-lib/fido2-net-lib/blob/f25c7e120b1d6b9ec6a1b7b4ca0310bd6d351ae9/Src/Fido2/DevelopmentInMemoryStore.cs#L75 

Storing the full date and timezone allows us to support web farms and region servers and translation to show the user if desired in browser's timezone will just work.  I know we can force utc time but with DatetimeOffset, it will automatically translate no matter how it was stored and sent.

joegoldman2:
Fixes #274


iamcarbon:
👍

aseigler (approved):
Looks good to me!

PR #438, opened 2023-10-10
Rename Fido2NetLib.cs file to Fido2.cs in order to match with the class name

joegoldman2:
Rename Fido2NetLib.cs to Fido2.cs in order to match with the class name


iamcarbon:
👍

aseigler (approved):

PR #442, opened 2023-10-11
Update to latest version of FIDO Metadata Service

joegoldman2:
Update to the latest MDS3 BLOB url. The format of the BLOB didn't change and the root certificate to validate it neither.


iamcarbon:
👍

aseigler (approved):
I think they ultimately point to the same place, but `https://mds3.fidoalliance.org/` is the correct currently published BLOB URL per https://fidoalliance.org/metadata/.

PR #447, opened 2023-10-14
`List<PublicKeyCredentialDescriptor>`/`IEnumerable<PublicKeyCredentialDescriptor>` to `IReadOnlyList<PublicKeyCredentialDescriptor>`

### LINKED ISSUE #191: Minor inconsistency between AllowCredentials and ExcludeCredentials
Hi, there is a tiny inconsistency in the API:
```cs
public class CredentialCreateOptions : Fido2ResponseBase
{
    ...
    public List<PublicKeyCredentialDescriptor> ExcludeCredentials { get; set; }
}

```
vs.
```cs
public class AssertionOptions : Fido2ResponseBase
{
    ...
    public IEnumerable<PublicKeyCredentialDescriptor> AllowCredentials { get; set; }
}
```

I therefore propose changing `AllowCredentials` to `List<PublicKeyCredentialDescriptor>`. This should not be a breaking change, as `List` is still `IEnumerable`.

joegoldman2:
Fixes #191.

The proposal in the issue was to change `AllowCredentials` to `List<PublicKeyCredentialDescriptor>`, I preferred to do the opposite because I think it's better to expose interfaces.

There are other parameters that deserve to be changed to `IEnumerable` as well, such as `storedDevicePublicKeys`: https://github.com/passwordless-lib/fido2-net-lib/blob/15263ca8bc4ceea84c3ef287300ac2962d9afe10/Src/Fido2/IFido2.cs#L20
I could do a PR for the others if this one is approved and merged.

### DIFF
--- a/BlazorWasmDemo/Client/Shared/UserService.cs
+++ b/BlazorWasmDemo/Client/Shared/UserService.cs
@@ -105,7 +105,7 @@ public class UserService
         {
             Console.WriteLine(e);
             var errorMessage = e.Message;
-            if (options.ExcludeCredentials?.Any() ?? false)
+            if (options.ExcludeCredentials?.Count > 0)
             {
                 errorMessage += " (You may have already registered this device)";
             }
--- a/BlazorWasmDemo/Client/Shared/UserService.cs
+++ b/BlazorWasmDemo/Client/Shared/UserService.cs
@@ -105,7 +105,7 @@ public class UserService
         {
             Console.WriteLine(e);
             var errorMessage = e.Message;
-            if (options.ExcludeCredentials?.Count() > 0)
+            if (options.ExcludeCredentials?.Any() ?? false)
             {
                 errorMessage += " (You may have already registered this device)";
             }
--- a/BlazorWasmDemo/Client/Shared/UserService.cs
+++ b/BlazorWasmDemo/Client/Shared/UserService.cs
@@ -105,7 +105,7 @@ public class UserService
         {
             Console.WriteLine(e);
             var errorMessage = e.Message;
-            if (options.ExcludeCredentials?.Count > 0)
+            if (options.ExcludeCredentials?.Count() > 0)
             {
                 errorMessage += " (You may have already registered this device)";
             }
--- a/Src/Fido2.Models/AssertionOptions.cs
+++ b/Src/Fido2.Models/AssertionOptions.cs
@@ -36,7 +36,7 @@ public class AssertionOptions : Fido2ResponseBase
     /// This OPTIONAL member contains a list of PublicKeyCredentialDescriptor objects representing public key credentials acceptable to the caller, in descending order of the caller’s preference(the first item in the list is the most preferred credential, and so on down the list)
     /// </summary>
     [JsonPropertyName("allowCredentials")]
-    public IEnumerable<PublicKeyCredentialDescriptor> AllowCredentials { get; set; }
+    public IReadOnlyList<PublicKeyCredentialDescriptor> AllowCredentials { get; set; } = new List<PublicKeyCredentialDescriptor>();
 
     /// <summary>
     /// This member describes the Relying Party's requirements regarding user verification for the get() operation. Eligible authenticators are filtered to only those capable of satisfying this requirement
@@ -51,7 +51,12 @@ public class AssertionOptions : Fido2ResponseBase
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public AuthenticationExtensionsClientInputs Extensions { get; set; }
 
-    public static AssertionOptions Create(Fido2Configuration config, byte[] challenge, IEnumerable<PublicKeyCredentialDescriptor> allowedCredentials, UserVerificationRequirement? userVerification, AuthenticationExtensionsClientInputs extensions)
+    public static AssertionOptions Create(
+        Fido2Configuration config,
+        byte[] challenge,
+        IReadOnlyList<PublicKeyCredentialDescriptor> allowedCredentials,
+        UserVerificationRequirement? userVerification,
+        AuthenticationExtensionsClientInputs extensions)
     {
         return new AssertionOptions()
         {
--- a/Src/Fido2.Models/AssertionOptions.cs
+++ b/Src/Fido2.Models/AssertionOptions.cs
@@ -36,7 +36,7 @@ public class AssertionOptions : Fido2ResponseBase
     /// This OPTIONAL member contains a list of PublicKeyCredentialDescriptor objects representing public key credentials acceptable to the caller, in descending order of the caller’s preference(the first item in the list is the most preferred credential, and so on down the list)
     /// </summary>
     [JsonPropertyName("allowCredentials")]
-    public IReadOnlyList<PublicKeyCredentialDescriptor> AllowCredentials { get; set; } = new List<PublicKeyCredentialDescriptor>();
+    public IReadOnlyList<PublicKeyCredentialDescriptor> AllowCredentials { get; set; } = Array.Empty<PublicKeyCredentialDescriptor>();
 
     /// <summary>
     /// This member describes the Relying Party's requirements regarding user verification for the get() operation. Eligible authenticators are filtered to only those capable of satisfying this requirement
@@ -65,7 +65,7 @@ public class AssertionOptions : Fido2ResponseBase
             Challenge = challenge,
             Timeout = config.Timeout,
             RpId = config.ServerDomain,
-            AllowCredentials = allowedCredentials ?? Array.Empty<PublicKeyCredentialDescriptor>(),
+            AllowCredentials = allowedCredentials,
             UserVerification = userVerification,
             Extensions = extensions
         };
--- a/Src/Fido2.Models/AssertionOptions.cs
+++ b/Src/Fido2.Models/AssertionOptions.cs
@@ -24,7 +24,7 @@ public class AssertionOptions : Fido2ResponseBase
     /// This member specifies a time, in milliseconds, that the caller is willing to wait for the call to complete. This is treated as a hint, and MAY be overridden by the client.
     /// </summary>
     [JsonPropertyName("timeout")]
-    public uint Timeout { get; set; }
+    public ulong Timeout { get; set; }
 
     /// <summary>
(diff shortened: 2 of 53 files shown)


Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -61,7 +61,7 @@ public sealed class CredentialCreateOptions : Fido2ResponseBase
     /// This member is intended for use by Relying Parties that wish to limit the creation of multiple credentials for the same account on a single authenticator.The client is requested to return an error if the new credential would be created on an authenticator that also contains one of the credentials enumerated in this parameter.
     /// </summary>
     [JsonPropertyName("excludeCredentials")]
-    public List<PublicKeyCredentialDescriptor> ExcludeCredentials { get; set; }
+    public IEnumerable<PublicKeyCredentialDescriptor> ExcludeCredentials { get; set; }
```
Regenhardt, line 64: There's supposed to be actual data here - how does a serializer know what to deserialise here? Do POCOs/DTOs really need interfaces on their data?

BlazorWasmDemo/Client/Shared/UserService.cs
```diff
@@ -105,7 +105,7 @@ public UserService(HttpClient httpClient, WebAuthn webAuthn)
         {
             Console.WriteLine(e);
             var errorMessage = e.Message;
-            if (options.ExcludeCredentials?.Count > 0)
+            if (options.ExcludeCredentials?.Count() > 0)
```
Regenhardt, line 108: Could use `.Any()` so it doesn't count all of the entries.

Src/Fido2.Models/CredentialCreateOptions.cs
dbeinder: Why not use IReadOnlyList<T> or IList<T> and still have an efficient .Count property and indexer if the goal is a common interface for lists and arrays?

I'd only use IEnumerable<T> to explicitly support lazy loading / streaming processing.

dbeinder: Sure, the library does not do anything with it, it just passes it through. But with that rationale one might as well advocate for using `object` ;)
The API consumer should also be considered, and they are being helped by having an abstraction that closely matches that data in the DTO. The data here will always be an array-style-thing and `IList` allows the API consumer to use their favorite flavor of that. `IEnumerable` is a magic generator box with potentially unbounded output, and requires the consumer to query element by element just to get a count. Even though we already know the actual object being transported has to be able to provide that information at near-zero cost.

Also from the perspective of minimizing surprises `IEnumerable` is questionable in this case. You remove `.ToList()` here:
```
var existingKeys = _demoStorage.GetCredentialsByUser(user).Select(c => c.Descriptor).ToList();
var options = _fido2.RequestNewCredential(user, existingKeys, ...);
return Json(options);
```
In a typical setup, the storage will be something like EntityFramework, and now `existingKeys` is a prepared SQL query. If there happens to be a problem with the database, the exception will be thrown in the bowels of the Json serializer, when the prepared query will finally be executed on the DB. And often controller functions don't even explicitly serialize, but returns the `options` objects directly leave that to the ASP.NET middleware, making for an even more interesting stack trace.

joegoldman2: > and requires the consumer to query element by element just to get a count.

The consumer, the library here, doesn't need to know the count or alter the data.

@abergs, @aseigler any thought?

dbeinder: The `CredentialCreateOptions` DTO is public API.
What would you call Blazor, if not an API consumer?

joegoldman2: Not sure to understand why you want to change it to something more specific than `IEnumerable<PublicKeyCredentialDescriptor>`.

The fido lib itself is not doing more than returning it,

https://github.com/passwordless-lib/fido2-net-lib/blob/eb312de4c580fb43c28f8e9928d0dcd6c1fec76d/Src/Fido2.Models/CredentialCreateOptions.cs#L99

and the value which will be serialised, or not, later.

It's the same thing for `AllowCredentials` in `AssertionOptions`

https://github.com/passwordless-lib/fido2-net-lib/blob/eb312de4c580fb43c28f8e9928d0dcd6c1fec76d/Src/Fido2.Models/AssertionOptions.cs#L39-L39

which is already a `IEnumerable`.

dbeinder: > Not sure to understand why you want to change it to something more specific
> The fido lib itself is not doing more than returning it, and the value which will be serialised, or not, later.

And a Blazor app may be deserializing it (into the same class) on the client side, and now has the inconvenience of `IEnumerable<T>`, instead of an array-like type. Which is why you had to change Blazor code for this PR.

And suggesting to users in the readme to pass in not-yet evaluated EF queries, which may only get executed later after the context has already been disposed is setting up another tripwire.

> It's the same thing for `AllowCredentials` in `AssertionOptions` which is already a `IEnumerable`

Which could be improved by changing that to `IList<T>` as suggested in #191

Regenhardt: I agree if I prefer IList (or anything collection-like) rather than IEnumerable, since it's definitely completed the moment I receive or pass it somewhere.

iamcarbon:
I would prefer IReadOnlyList<T> to IEnumerable<T> here.

joegoldman2:
> I would prefer IReadOnlyList to IEnumerable here.

Ok I will do the change. Do you want me to change also `AllowCredentials` in `AssertionOptions`?

Src/Fido2.Models/CredentialCreateOptions.cs
joegoldman2: Changed to `IReadOnlyList`. Demo app no longer impacted by this change 👍

joegoldman2:
> I would prefer IReadOnlyList to IEnumerable here.

Done! This PR is ready to be reviewed.

Src/Fido2.Models/AssertionOptions.cs
```diff
@@ -36,7 +36,7 @@ public class AssertionOptions : Fido2ResponseBase
     /// This OPTIONAL member contains a list of PublicKeyCredentialDescriptor objects representing public key credentials acceptable to the caller, in descending order of the caller’s preference(the first item in the list is the most preferred credential, and so on down the list)
     /// </summary>
     [JsonPropertyName("allowCredentials")]
-    public IEnumerable<PublicKeyCredentialDescriptor> AllowCredentials { get; set; }
+    public IReadOnlyList<PublicKeyCredentialDescriptor> AllowCredentials { get; set; } = new List<PublicKeyCredentialDescriptor>();
```
abergs, line 39: @iamcarbon a.k.a Mr Zero Allocations, do we want to leave this as null? And if we do want to set a value, should it be something more like Array.Empty?

Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -61,7 +61,7 @@ public sealed class CredentialCreateOptions : Fido2ResponseBase
     /// This member is intended for use by Relying Parties that wish to limit the creation of multiple credentials for the same account on a single authenticator.The client is requested to return an error if the new credential would be created on an authenticator that also contains one of the credentials enumerated in this parameter.
     /// </summary>
     [JsonPropertyName("excludeCredentials")]
-    public List<PublicKeyCredentialDescriptor> ExcludeCredentials { get; set; }
+    public IReadOnlyList<PublicKeyCredentialDescriptor> ExcludeCredentials { get; set; } = new List<PublicKeyCredentialDescriptor>();
```
abergs, line 64: Same observation

Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -96,7 +103,7 @@ public static CredentialCreateOptions Create(Fido2Configuration config, byte[] c
             },
             AuthenticatorSelection = authenticatorSelection,
             Attestation = attestationConveyancePreference,
-            ExcludeCredentials = excludeCredentials ?? new List<PublicKeyCredentialDescriptor>(),
+            ExcludeCredentials = excludeCredentials ?? Array.Empty<PublicKeyCredentialDescriptor>(),
```
abergs, line 106: Same here actually, spec says this is optional: https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-excludecredentials

abergs (commented):
Aside from my question about not using `null` or different init values, It's Approved 👍

Src/Fido2.Models/CredentialCreateOptions.cs
Regenhardt: The spec does say ` defaulting to []`, which would fit the general meaning of "null == no information, empty list == we know that no things exist". Were it not for that part in the spec though, I would have no preference either way.

abergs: @Regenhardt The way I read the spec, it just means that we don't need to set it. If not set, the browser will default it to be an empty array that is passed to the Authenticator. Not that the RP (we) is required to set it to a empty array.

Src/Fido2.Models/AssertionOptions.cs
joegoldman2: I agree with you, this allocation is not necessary. I replaced it `Array.Empty`.

Src/Fido2.Models/CredentialCreateOptions.cs
joegoldman2: Same as above. Replaced by `Array.Empty`.

Src/Fido2.Models/CredentialCreateOptions.cs
joegoldman2: Agree with you, the default value is not necessary it. I removed it and tested manually the change.

abergs:
@iamcarbon Do you have any comments regarding the allocations?

joegoldman2:
> @iamcarbon Do you have any comments regarding the allocations?

@abergs There is no additional allocation in the current PR. Only a default value to `Array.Empty` that can be discussed if you want.

joegoldman2:
@iamcarbon any thoughts?

iamcarbon:
Looks good to me! 👍

PR #448, opened 2023-10-14
[Draft] Add credProtect extension to Fido2.Models

dbeinder:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API/WebAuthn_extensions#credprotect
https://fidoalliance.org/specs/fido-v2.2-rd-20230321/fido-client-to-authenticator-protocol-v2.2-rd-20230321.html#sctn-credProtect-extension

The input part should be okay, but does the extension output need parsing from JSON? Or should it be extracted from the `authenticator data` CBOR?


jonashendrickx:
Confirmed input to be working.

aseigler (approved):
LGTM!

abergs (approved):

PR #450, opened 2023-10-23
Rename `clientDataJson` to `clientDataJSON`

### LINKED ISSUE #338: AuthenticatorAttestationRawResponse attestationResponse getting null.
Hi,

I had taken a reference from sample demo for  passwordless. 

where we have a method called MakeCredential

 [HttpPost]
 [Route("/makeCredential")]
 public async Task<JsonResult> MakeCredential([FromBody] AuthenticatorAttestationRawResponse attestationResponse, CancellationToken cancellationToken)
 
 In demo we do get the attestationResponse value but when we try to implement the same thing in our project,
 the response which we get  for attestationResponse is null although when checked in network tab the payload is passed.
 Below is the screen shot of payload which we are passing.
 
 Could you please help us to figure out this issue?
 

image

joegoldman2:
Fixes #338.

### DIFF
--- a/BlazorWasmDemo/Server/Controllers/UserController.cs
+++ b/BlazorWasmDemo/Server/Controllers/UserController.cs
@@ -173,7 +173,7 @@ public class UserController : ControllerBase
                 IsBackupEligible = result.Result.IsBackupEligible,
                 IsBackedUp = result.Result.IsBackedUp,
                 AttestationObject = result.Result.AttestationObject,
-                AttestationClientDataJSON = result.Result.AttestationClientDataJson,
+                AttestationClientDataJson = result.Result.AttestationClientDataJson,
             });
 
             // 5. Now we need to remove the options from the pending dictionary
--- a/Demo/Controller.cs
+++ b/Demo/Controller.cs
@@ -131,7 +131,7 @@ public class MyController : Controller
                 IsBackupEligible = success.Result.IsBackupEligible,
                 IsBackedUp = success.Result.IsBackedUp,
                 AttestationObject = success.Result.AttestationObject,
-                AttestationClientDataJSON = success.Result.AttestationClientDataJson,
+                AttestationClientDataJson = success.Result.AttestationClientDataJson,
                 DevicePublicKeys = new List<byte[]>() { success.Result.DevicePublicKey }
             });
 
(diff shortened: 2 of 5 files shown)


joegoldman2:
@abergs, @aseigler this PR is ready to be reviewed.

aseigler:
This is one of those things that has been inconsistent from day one.  Should we take the call to rename all instances of `clientDataJson` to `clientDataJSON` and not just these two?

aseigler:
> > This is one of those things that has been inconsistent from day one. Should we take the call to rename all instances of `clientDataJson` to `clientDataJSON` and not just these two?
> 
> I was not sure if we want to use `clientDataJSON` with uppercase also in the c#. If you are ok with this, I can rename the property also in the c# contract.

@abergs , @iamcarbon , opinions?

iamcarbon:
My personal preference would be to use clientDataJSON during JSON serialization, and rename all the C# properties to "ClientDataJson" to remain consistent with .NET conventions.

joegoldman2:
> My personal preference would be to use clientDataJSON during JSON serialization, and rename all the C# properties to "ClientDataJson" to remain consistent with .NET conventions.

So it's normally already the case:

https://github.com/passwordless-lib/fido2-net-lib/blob/11c78ce956ec2552bf122e2e14281080b5a8374e/Src/Fido2.Models/AuthenticatorAssertionRawResponse.cs#L43-L44
https://github.com/passwordless-lib/fido2-net-lib/blob/11c78ce956ec2552bf122e2e14281080b5a8374e/Src/Fido2.Models/AuthenticatorAttestationRawResponse.cs#L33-L34

abergs:
For serialization/deserialization it should definitely be named `clientDataJSON`. That is what is used [in the spec](https://www.w3.org/TR/webauthn-3/#iface-authenticatorresponse).

I would have tried to make the point on keeping `JSON` in .NET to make it less error-prone, but we rely on `[JsonPropertyName]` for many other fields, so that point is mute.

Go ahead and rename it to `Json` in .net @joegoldman2.

abergs (changes requested):
As discussed

joegoldman2:
> Go ahead and rename it to `Json` in .net @joegoldman2.

Thank you for the review. It's done.

abergs (approved):
Thanks, looks good!

PR #453, opened 2023-10-23
Map authenticator transports on server side

joegoldman2:
Map the authenticator transports on server side. The demo app already stores the value:

https://github.com/passwordless-lib/fido2-net-lib/blob/ebfd75d00a945a80578b27a256223122cc2aefe9/Demo/Controller.cs#L130

### DIFF
--- a/Demo/wwwroot/js/custom.register.js
+++ b/Demo/wwwroot/js/custom.register.js
@@ -125,8 +125,7 @@ async function registerNewCredential(newCredential) {
         extensions: newCredential.getClientExtensionResults(),
         response: {
             AttestationObject: coerceToBase64Url(attestationObject),
-            clientDataJSON: coerceToBase64Url(clientDataJSON),
-            transports: newCredential.response.getTransports(),
+            clientDataJSON: coerceToBase64Url(clientDataJSON)
         },
     };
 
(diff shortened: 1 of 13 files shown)


joegoldman2:
@abergs, @aseigler this PR is ready to be reviewed

abergs (approved):

abergs (changes requested):
Actually - [when reading the spec](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialdescriptor-transports), My opinion is that we should add transports to our server side raw class.

joegoldman2:
@abergs I added the transports on server side and tested manually (in addition to updating the unit tests).

abergs (approved):
@joegoldman2 Thank you, I think this looks solid.

Would welcome any second opinions from @iamcarbon or @aseigler, but I'm approving this.

joegoldman2:
@iamcarbon, @aseigler any thoughs please?

aseigler:
Code looks reasonable to me, not sure what is going on with those failed tests.  Is that new or an environment issue?

joegoldman2:
> Code looks reasonable to me, not sure what is going on with those failed tests. Is that new or an environment issue?

The dotnet format step has failed with `Unable to locate dotnet CLI. Ensure that it is on the PATH.`. Are you able to re-run the failed job?

abergs:
I'm trying to re-run those checks to see if can get this in!

PR #465, opened 2023-11-17
Remove binary serialization

iamcarbon:
Binary serialization support is no longer recommended as .NET5.0 and is obsoleted entirely in .NET 8.0 in preparation for its removal in .NET 9.0.

https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md

This PR removes binary serialization support on Exceptions (and eliminates the warnings when targeting .NET8.0) -- and helps ensure no-one relies on these when support is removed in .NET 9.0.

### DIFF
--- a/Src/Fido2.Models/Exceptions/Fido2MetadataException.cs
+++ b/Src/Fido2.Models/Exceptions/Fido2MetadataException.cs
@@ -1,9 +1,5 @@
-﻿using System;
-using System.Runtime.Serialization;
+﻿namespace Fido2NetLib;
 
-namespace Fido2NetLib;
-
-[Serializable]
 public class Fido2MetadataException : Exception
 {
     public Fido2MetadataException()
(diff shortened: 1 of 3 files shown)


iamcarbon:
@abergs Ready for review.

abergs:
@iamcarbon what side effects would this cause?

iamcarbon:
This could affect users that utilize a legacy library that serializes exceptions using the BinaryFormatter (e.g. an old logging framework) -- which were more common on .NET Framework 10 years ago.

.NET core stopped introducing new serializable exceptions, so it's very unlikely a modern .NET would be relying on binary serialization -- or using any libraries that utilize binary serialization. 

If someone happens to be using binary serialization on modern .NET, they should be migrating away ASAP.

Screenshot 2023-11-22 at 10 48 56 AM


https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=net-8.0

iamcarbon:
@aseigler Let me know if you see any reason not keep these around. These attributes cause an error on .NET 8.0 and will be need to be suppressed if we want to continue to support them on .NET 6.0.

PR #474, opened 2023-12-21
Rename 'extensions' to 'clientExtensionResults' for deserialization

jonashendrickx:
We are currently using 'extensions' instead of 'clientExtensionResults', this requires us to map the property from the PublicKeyCredential object in the browser, or Google FIDO2 API.

- https://w3c.github.io/webauthn/#iface-pkcredential

The demo applications have been left to use the original 'extensions' property for the registration, to demonstrate the change is backwards compatible.

I first renamed the property, and ensured the project was building and all references were correctly pointing to the new property. Then added the obsolete property back. There shouldn't be any leaks.


abergs (approved):
LGTM. Leaving some room for @aseigler to comment but will merge later if there are not comments.

shubhamji88:
We found this issue while using the Google android FIDO2 API. Currently, we have fixed this issue [here](https://github.com/bitwarden/passwordless-android/blob/main/passwordless/src/main/java/dev/passwordless/android/rest/converters/PublicKeyCredentialConverter.kt) by just renaming the "clientExtensionResults" to "extensions" and then sending to backend.

abergs:
@jonashendrickx This PR contains conflicts that stops it from being merged

PR #524, opened 2024-07-06
Implement PublicKeyCredentialHint for WebAuthn L3

jonashendrickx:


Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -55,6 +55,34 @@ public sealed class CredentialCreateOptions : Fido2ResponseBase
...
+        get
+        {
+            return _hints;
+        }
+        set
+        {
+            if (value != null && value.Any())
+            {
+                AuthenticatorSelection.AuthenticatorAttachment ??= new AuthenticatorAttachment();
+                AuthenticatorSelection.AuthenticatorAttachment = AuthenticatorAttachment.CrossPlatform;
```
abergs, line 79: I don't think this is correct, see the spec:
CleanShot 2024-07-11 at 14 06 51@2x

abergs (changes requested):
Thanks! Looks mostly good, see comment on AuthenticatorAttachment.

Also - this PR is missing the implementation on assertion - hints are supported there too, not only on create.

Src/Fido2.Models/CredentialCreateOptions.cs
jonashendrickx: @abergs Ah right the second one has to be `platform`. I completely missed that.

jonashendrickx: @abergs Ready.

Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -55,6 +55,52 @@ public sealed class CredentialCreateOptions : Fido2ResponseBase
...
+        {
+            if (value.Any())
+            {
+                switch (value[0])
+                {
+                    case PublicKeyCredentialHint.SecurityKey:
+                    case PublicKeyCredentialHint.Hybrid:
+                        {
+                            AuthenticatorSelection.AuthenticatorAttachment ??= new AuthenticatorAttachment();
+                            AuthenticatorSelection.AuthenticatorAttachment = AuthenticatorAttachment.CrossPlatform;
```
abergs, line 89: @jonashendrickx Maybe I'm missing something, but what's the reason for double assignment to `AuthenticatorSelection.AuthenticatorAttachment`?

abergs, line 89:
```suggestion
                            AuthenticatorSelection.AuthenticatorAttachment = AuthenticatorAttachment.CrossPlatform;
```

jonashendrickx: @abergs It had to be `AuthenticatorSelection` which I had to initialize, as it may be null.

> authenticatorSelection, of type [AuthenticatorSelectionCriteria](https://www.w3.org/TR/webauthn-3/#dictdef-authenticatorselectioncriteria)
> The [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) MAY use this OPTIONAL member to specify capabilities and settings that the [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) MUST or SHOULD satisfy to participate in the [create()](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create) operation. See [§ 5.4.4 Authenticator Selection Criteria (dictionary AuthenticatorSelectionCriteria)](https://www.w3.org/TR/webauthn-3/#dictionary-authenticatorSelection).

abergs (approved):
Thanks for the fix, this makes sense!

PR #528, opened 2024-07-15
Resolve StoredCredential.Descriptor at access(?)

abergs:
While we might want to cache it, this is to simplify the concept so that devs more easier understand how the piecies fit toghether"

### DIFF
--- a/Src/Fido2.Development/StoredCredential.cs
+++ b/Src/Fido2.Development/StoredCredential.cs
@@ -52,6 +52,9 @@ public class StoredCredential
 
     public byte[] UserId { get; set; }
 
+    /// <summary>
+    /// Exposes an Descriptor Object for this credential, used as input to the library for certain operations.
+    /// </summary>
     public PublicKeyCredentialDescriptor Descriptor => new(PublicKeyCredentialType.PublicKey, Id, Transports);
 
     public byte[] UserHandle { get; set; }
--- a/Src/Fido2.Development/StoredCredential.cs
+++ b/Src/Fido2.Development/StoredCredential.cs
@@ -9,7 +9,7 @@ public class StoredCredential
     /// <summary>
     /// The Credential ID of the public key credential source.
     /// </summary>
-    public byte[] Id { get; set; }
+    public required byte[] Id { get; set; }
 
     /// <summary>
     /// The credential public key of the public key credential source.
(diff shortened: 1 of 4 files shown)


Src/Fido2.Development/StoredCredential.cs
```diff
@@ -52,7 +52,7 @@ public class StoredCredential
 
     public byte[] UserId { get; set; }
 
-    public PublicKeyCredentialDescriptor Descriptor { get; set; }
```
abergs, line 55: For less allocations we could cache the instance of the PublicKeyCredentialDescriptor as a private member.

Src/Fido2.Development/StoredCredential.cs
```diff
@@ -52,7 +52,7 @@ public class StoredCredential
...
+    public PublicKeyCredentialDescriptor Descriptor => new(PublicKeyCredentialType.PublicKey, Id, Transports);
```
abergs, line 55: For less allocations we could cache the instance of the PublicKeyCredentialDescriptor as a private member.

abergs:
@Regenhardt Does this perhaps strike a good balance on your original concerns in #510

Regenhardt:
Thanks for the ping, yes this is a nice solution for the redundancy. Could use an XML summary on the descriptor and maybe on the Id mentioning it as being the actual public key or something.

Src/Fido2.Development/StoredCredential.cs
```diff
@@ -9,7 +9,7 @@ public class StoredCredential
     /// <summary>
     /// The Credential ID of the public key credential source.
     /// </summary>
-    public byte[] Id { get; set; }
+    public required byte[] Id { get; set; }
```
iamcarbon, line 12: Very happy to see we're on .NET8, and are able to use a required property here.

iamcarbon: Can this also be made immutable with { get; init; }?

Src/Fido2.Development/StoredCredential.cs
iamcarbon: This looks good as-is to me. The consumer is likely to capture this into a variable, and it's a tiny object.

Src/Fido2.Development/StoredCredential.cs
abergs: @iamcarbon It probably could, but that opens the question if not more of these should be made requried/init-only etc so in an effort to merge this I won't add it.

Also - I need to remind myself that this type is only used in the Development namespace, so it's just used for demoes etc.d

PR #529, opened 2024-07-15
Remove base-class, Status and ErrorMessage

abergs:


Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
```diff
@@ -5,16 +5,9 @@ namespace Fido2NetLib.Objects;
...
 {
-    public MakeNewCredentialResult(string status, string errorMessage, RegisteredPublicKeyCredential? result)
-    {
-        Status = status;
-        ErrorMessage = errorMessage;
-        Result = result;
-    }
-
-    public RegisteredPublicKeyCredential? Result { get; }
+    public RegisteredPublicKeyCredential Credential { get; } = credential;
```
abergs, line 10: I also took the opportunity to `.Result` to `.Credential`.

We could go even further and remove this class and just immediately return the credential -- that is more aligned to the Options-methods that immediately return Options.

My only concern is that this would generate even more of a breaking changes as well as make it hard to introduce/return any information that is not part of what we call the `RegisteredPublicKeyCredential`

iamcarbon:
@abergs I had also explored removing the base class during the initial refactoring -- and agree this is a good time to take the break and move these values to the endpoints.

Src/Fido2.Models/Objects/MakeNewCredentialResult.cs
iamcarbon: + 1 for returning the credential directly and extending RegisteredPublicKeyCredential as needed.

abergs: Alright, that was all the encouragement I needed. It's been done.

abergs:
Alright, I was able to run the conformance tests on this branch and it went fine. Will merge once tests are OK.

PR #456, opened 2023-10-24
FIDO Conformance Tools v1.7.15 fixes

googyi:
passes: 128 failures: 19 
->
passes:164 failures: 0

changes / fixed testcases

TrustAnchor.cs : 32
Server-ServerAuthenticatorAttestationResponse-Resp-5 Test server processing "packed" FULL attestation F-10 Send ServerAuthenticatorAttestationResponse with FULL "packed" attestation, with attStmt.x5c containing full chain, and check that server returns an error https://datatracker.ietf.org/doc/html/rfc5280#section-6.1

AuthenticatorAttestationRawResponse.cs : 18
Server-ServerAuthenticatorAttestationResponse-Resp-1 Test server processing ServerAuthenticatorAttestationResponse structure F-4 Send ServerAuthenticatorAttestationResponse that is missing "type" field and check that server returns an error

CredentialCreateOptions.cs : 96
Server-ServerAuthenticatorAttestationResponse-Resp-4 Test server support of the authentication algorithms P-8 Send a valid ServerAuthenticatorAttestationResponse with SELF "packed" attestation, for "ALG_SIGN_RSASSA_PKCSV15_SHA1_RAW" aka "RS1" algorithm, and check that server succeeds Server-ServerAuthenticatorAttestationResponse-Resp-9 Test server processing "tpm" attestation P-2 Send a valid ServerAuthenticatorAttestationResponse with "tpm" attestation for SHA-1, and check that server succeeds

CredentialCreateOptions.cs  : 210
Server-ServerPublicKeyCredentialCreationOptions-Req-1 Test server generating ServerPublicKeyCredentialCreationOptionsRequest P-1 Get ServerPublicKeyCredentialCreationOptionsResponse, and check that: (a) response MUST contain ...

AuthenticationExtensionsClientInputs.cs : 23 public string AppID { private get; set; } Server-ServerPublicKeyCredentialGetOptionsResponse-Req-1 Test server generating ServerPublicKeyCredentialGetOptionsResponse P-1 Get ServerPublicKeyCredentialGetOptionsResponse, and check that: (a) response MUST contain ...

AuthenticationExtensionsClientInputs.cs :  44 public bool? UserVerificationMethod { private get; set; } Server-ServerPublicKeyCredentialGetOptionsResponse-Req-1 Test server generating ServerPublicKeyCredentialGetOptionsResponse P-1 Get ServerPublicKeyCredentialGetOptionsResponse, and check that: (a) response MUST contain ...

[...]

PR #531, opened 2024-07-16
conformance 1.7.20 4

abergs:
This is based on #456 and results in a 100% pass on the v1.7.20-4 tool.

image

- **FIDO Conformance Tools v1.7.15 fixes**
- **Json serialization fix**
- **Unit test fix**
- **tokenbindig, AppId, UVP**
- **unit test fix (tokenbinding dto parsing)**
- **fix azure pipeline**
- **Improve trustanchor test coverage**
- **TestPackedttestationAsyncFailTrustAnchorOnRootCertInTrustPath only works on Windows**
- **Do not make this private**
- **Keep Tokenbinding around**
- **Update AuthenticatorAssertionResponse.cs**
- **Added XML comments to requestTokenBinding**

### DIFF
--- a/Src/Fido2.Models/CredentialCreateOptions.cs
+++ b/Src/Fido2.Models/CredentialCreateOptions.cs
@@ -241,7 +241,7 @@ public class AuthenticatorSelection
     [JsonPropertyName("residentKey")]
     public ResidentKeyRequirement ResidentKey
     {
-        private get => _residentKey;
+        get => _residentKey;
         set
         {
             _residentKey = value;
--- a/Src/Fido2.Models/CredentialCreateOptions.cs
+++ b/Src/Fido2.Models/CredentialCreateOptions.cs
@@ -134,6 +134,7 @@ public sealed class CredentialCreateOptions
                 PubKeyCredParam.ES512,
                 PubKeyCredParam.RS512,
                 PubKeyCredParam.PS512,
+                PubKeyCredParam.RS1
             ],
             AuthenticatorSelection = authenticatorSelection,
             Attestation = attestationConveyancePreference,
@@ -185,6 +186,7 @@ public sealed class PubKeyCredParam(
     public static readonly PubKeyCredParam PS384 = new(COSE.Algorithm.PS384);
     public static readonly PubKeyCredParam PS512 = new(COSE.Algorithm.PS512);
     public static readonly PubKeyCredParam Ed25519 = new(COSE.Algorithm.EdDSA);
+    public static readonly PubKeyCredParam RS1 = new(COSE.Algorithm.RS1);
 }
 
 /// <summary>
@@ -239,7 +241,7 @@ public class AuthenticatorSelection
     [JsonPropertyName("residentKey")]
     public ResidentKeyRequirement ResidentKey
     {
-        get => _residentKey;
+        private get => _residentKey;
         set
         {
             _residentKey = value;
--- a/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
+++ b/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
@@ -19,13 +19,8 @@ public sealed class AuthenticationExtensionsClientInputs
     /// https://www.w3.org/TR/webauthn/#sctn-appid-extension
     /// </summary>
     [JsonPropertyName("appid")]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string AppID { private get; set; }
-
-    public string GetAppID()
-    {
-        return AppID;
-    }
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
+    public string AppID { get; set; }
 
     /// <summary>
     /// This extension enables the WebAuthn Relying Party to determine which extensions the authenticator supports.
--- a/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
+++ b/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
@@ -12,7 +12,7 @@ public sealed class AuthenticationExtensionsClientInputs
     /// </summary>
     [JsonPropertyName("example.extension.bool")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public object Example { get; set; }
+    public bool? Example { get; set; }
 
     /// <summary>
     /// This extension allows WebAuthn Relying Parties that have previously registered a credential using the legacy FIDO JavaScript APIs to request an assertion.
--- a/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
+++ b/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
@@ -33,6 +33,7 @@ public sealed class AuthenticationExtensionsClientInputs
     /// <summary>
     /// This extension enables use of a user verification method.
     /// https://www.w3.org/TR/webauthn/#sctn-uvm-extension
+    /// TODO: Remove this completely as it's removed in L3
     /// </summary>
     [JsonPropertyName("uvm")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
--- a/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
+++ b/Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
@@ -20,7 +20,12 @@ public sealed class AuthenticationExtensionsClientInputs
     /// </summary>
     [JsonPropertyName("appid")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string AppID { get; set; }
+    public string AppID { private get; set; }
+
+    public string GetAppID()
+    {
+        return AppID;
+    }
 
     /// <summary>
     /// This extension enables the WebAuthn Relying Party to determine which extensions the authenticator supports.
@@ -36,7 +41,7 @@ public sealed class AuthenticationExtensionsClientInputs
     /// </summary>
     [JsonPropertyName("uvm")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public bool? UserVerificationMethod { get; set; }
+    public bool? UserVerificationMethod { private get; set; }
 
 #nullable enable
     /// <summary>
--- a/Src/Fido2/AuthenticatorAssertionResponse.cs
+++ b/Src/Fido2/AuthenticatorAssertionResponse.cs
@@ -127,7 +127,7 @@ public sealed class AuthenticatorAssertionResponse : AuthenticatorResponse
             throw new Fido2VerificationException(Fido2ErrorCode.InvalidRpidHash, Fido2ErrorMessages.InvalidRpidHash);
 
         var conformanceTesting = metadataService != null && metadataService.ConformanceTesting();
-        
+
         // 14. Verify that the UP bit of the flags in authData is set.
         // Todo: Conformance testing verifies the UVP flags differently than W3C spec, simplify this by removing the mention of conformanceTesting when conformance tools are updated) 
         if (!authData.UserPresent && !conformanceTesting)
@@ -136,7 +136,7 @@ public sealed class AuthenticatorAssertionResponse : AuthenticatorResponse
         // 15. If the Relying Party requires user verification for this assertion, verify that the UV bit of the flags in authData is set.
         if (options.UserVerification is UserVerificationRequirement.Required && !authData.UserVerified)
             throw new Fido2VerificationException(Fido2ErrorCode.UserVerificationRequirementNotMet, Fido2ErrorMessages.UserVerificationRequirementNotMet);
-        
+
         // 16. If the credential backup state is used as part of Relying Party business logic or policy, let currentBe and currentBs be the values of the BE and BS bits, respectively, of the flags in authData.
         // Compare currentBe and currentBs with credentialRecord.BE and credentialRecord.BS and apply Relying Party policy, if any.
         if (authData.IsBackupEligible && config.BackupEligibleCredentialPolicy is Fido2Configuration.CredentialBackupPolicy.Disallowed ||
--- a/Src/Fido2/AuthenticatorAssertionResponse.cs
+++ b/Src/Fido2/AuthenticatorAssertionResponse.cs
@@ -61,9 +61,10 @@ public sealed class AuthenticatorAssertionResponse : AuthenticatorResponse
         uint storedSignatureCounter,
         IsUserHandleOwnerOfCredentialIdAsync isUserHandleOwnerOfCredId,
         IMetadataService? metadataService,
+        byte[]? requestTokenBindingId,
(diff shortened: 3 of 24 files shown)


Src/Fido2.Models/CredentialCreateOptions.cs
```diff
@@ -136,6 +136,7 @@ public IReadOnlyList<PublicKeyCredentialHint> Hints
                 PubKeyCredParam.ES512,
                 PubKeyCredParam.RS512,
                 PubKeyCredParam.PS512,
+                PubKeyCredParam.RS1
```
abergs, line 139: P-8 Send a valid ServerAuthenticatorAttestationResponse with SELF "packed" attestation, for "ALG_SIGN_RSASSA_PKCSV15_SHA1_RAW" aka "RS1" algorithm, and check that server succeeds‣

Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
```diff
@@ -12,14 +12,14 @@ public sealed class AuthenticationExtensionsClientInputs
     /// </summary>
     [JsonPropertyName("example.extension.bool")]
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public object Example { get; set; }
```
abergs, line 15: JSON Serialization fails if this is object.

Src/Fido2.Models/Objects/AuthenticationExtensionsClientInputs.cs
```diff
@@ -12,14 +12,14 @@ public sealed class AuthenticationExtensionsClientInputs
...
+    public bool? Example { get; set; }
 
     /// <summary>
     /// This extension allows WebAuthn Relying Parties that have previously registered a credential using the legacy FIDO JavaScript APIs to request an assertion.
     /// https://www.w3.org/TR/webauthn/#sctn-appid-extension
     /// </summary>
     [JsonPropertyName("appid")]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
     public string AppID { get; set; }
```
abergs, line 23: CleanShot 2024-07-16 at 10 24 02

abergs: Conformance tests do not want AppId to be returned.

Src/Fido2/AuthenticatorAssertionResponse.cs
```diff
@@ -61,9 +62,10 @@ public static AuthenticatorAssertionResponse Parse(AuthenticatorAssertionRawResp
         uint storedSignatureCounter,
         IsUserHandleOwnerOfCredentialIdAsync isUserHandleOwnerOfCredId,
         IMetadataService? metadataService,
+        byte[]? requestTokenBindingId,
         CancellationToken cancellationToken = default)
     {
-        BaseVerify(config.FullyQualifiedOrigins, options.Challenge);
+        BaseVerify(config.FullyQualifiedOrigins, options.Challenge, requestTokenBindingId);
```
abergs, line 68: Conformance requires requestTokenBinding to still be around, should be removed once the tool is updated.

Src/Fido2/Extensions/CryptoUtils.cs
```diff
@@ -59,7 +59,9 @@ public static bool ValidateTrustChain(X509Certificate2[] trustPath, X509Certific
         // A trust anchor can be a root certificate, an intermediate CA certificate or even the attestation certificate itself.
 
         // Let's check the simplest case first.  If subject and issuer are the same, and the attestation cert is in the list, that's all the validation we need
-        if (trustPath.Length == 1 && trustPath[0].Subject.Equals(trustPath[0].Issuer, StringComparison.Ordinal))
+
+        // We have the same singular root cert in trustpath and it is in attestationRootCertificates
+        if (trustPath.Length == 1)
```
abergs, line 64: @aseigler Is it really ok to *not* equality compare the path?

Without this we get: 
CleanShot 2024-07-16 at 11 23 02

I could throw in a `|| conformanceTesting`-related check

Src/Fido2/TrustAnchor.cs
```diff
@@ -1,60 +1,65 @@
...
+
+                for (int i = 0; i < attestationRootCertificates.Length; i++)
+                {
+                    attestationRootCertificates[i] = new X509Certificate2(Convert.FromBase64String(certStrings[i]));
+                }
+
+                if (trustPath.Length > 1 && attestationRootCertificates.Any(c => string.Equals(c.Thumbprint, trustPath[^1].Thumbprint, StringComparison.Ordinal)))
+                {
+                    throw new Fido2VerificationException(Fido2ErrorMessages.InvalidCertificateChain);
+                }
```
abergs, line 35: @aseigler Would appreciate your eyes here too, but to me it just looks like we have the addition of above if-block that is more strict than without it.

Test/CryptoUtilsTests.cs
```diff
@@ -66,8 +66,8 @@ public void TestValidateTrustChainSubAnchor()
 
         Assert.False(0 == attestationRootCertificates[0].Issuer.CompareTo(attestationRootCertificates[0].Subject));
         Assert.True(CryptoUtils.ValidateTrustChain(trustPath, attestationRootCertificates));
-        Assert.False(CryptoUtils.ValidateTrustChain(trustPath, trustPath));
-        Assert.False(CryptoUtils.ValidateTrustChain(attestationRootCertificates, attestationRootCertificates));
+        Assert.True(CryptoUtils.ValidateTrustChain(trustPath, trustPath));
+        Assert.True(CryptoUtils.ValidateTrustChain(attestationRootCertificates, attestationRootCertificates));
```
abergs, line 70: These test changes are also something I wonder about @aseigler

abergs:
@aseigler See mentions. If I don't get your approval I might move those changes to a secondary PR wand wait for your review, while still landing some of these fixes.

Src/Fido2.Models/CredentialCreateOptions.cs
aseigler: Ideally, this should be made configurable instead of being hard coded, with defaults excluding RS1 (IMHO).

Src/Fido2/Extensions/CryptoUtils.cs
aseigler: The check to see if the subject and the issuer are the same is based on logic described in https://datatracker.ietf.org/doc/html/rfc5280#section-3.2.  I thought we had already covered this before, where the batch certificate referenced in the metadata is not a root...I'll have to review this more thoroughly.

Src/Fido2.Models/CredentialCreateOptions.cs
abergs: Tracked in #533, resolving this conversation.

Src/Fido2/IFido2.cs
```diff
@@ -20,12 +20,14 @@ Task<VerifyAssertionResult> MakeAssertionAsync(
...
         IsUserHandleOwnerOfCredentialIdAsync isUserHandleOwnerOfCredentialIdCallback,
+        byte[]? requestTokenBindingId = null,
         CancellationToken cancellationToken = default);
 
     Task<RegisteredPublicKeyCredential> MakeNewCredentialAsync(
         AuthenticatorAttestationRawResponse attestationResponse,
         CredentialCreateOptions originalOptions,
         IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser,
+        byte[]? requestTokenBindingId = null,
         CancellationToken cancellationToken = default);
```
iamcarbon, line 31: Consider making a new object for the parameters so we can continue to add or remove parameters in the future without a binary breaking change.

```csharp
public class MakeNewCredentialRequest
{    
    public required AuthenticatorAttestationRawResponse AttestationResponse { get; init; }
    
    public required CredentialCreateOptions OriginalOptions { get; init; }
    
    // for Fido2 2024 conformance testing only
    [EditorBrowsable(EditorBrowsableState.Never)] 
    public byte[]? RequestTokenBindingId { get; init; }
}
```

Src/Fido2/TrustAnchor.cs
```diff
@@ -1,60 +1,65 @@
...
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+
+using Fido2NetLib.Exceptions;
+
+namespace Fido2NetLib;
+
+public static class TrustAnchor
+{
+    public static void Verify(MetadataBLOBPayloadEntry? metadataEntry, X509Certificate2[] trustPath, bool conformance)
```
iamcarbon, line 11: consider:

```
public static void Verify(MetadataBLOBPayloadEntry? metadataEntry, X509Certificate2[] trustPath, FidoValidationMode validationMode =  FidoValidationMode.Default)
```

iamcarbon (changes requested):

Src/Fido2/Extensions/CryptoUtils.cs
iamcarbon: @abergs I saw that we removed the check to make sure the trustPath's root subject is equal to the issuer. 

What does the conformance test use for the Subject and Issuer (where they don't match). Is there a case, in production, where these values would also not match? Are they completely different values, or just encoded differently?

Src/Fido2/TrustAnchor.cs
iamcarbon: @abergs Are we able to reference the rule in https://datatracker.ietf.org/doc/html/rfc5280 that this new condition enforces?

Src/Fido2/Fido2.cs
```diff
@@ -65,16 +65,18 @@ public CredentialCreateOptions RequestNewCredential(
...
     /// <param name="originalOptions">The original options that was sent to the client.</param>
     /// <param name="isCredentialIdUniqueToUser">The delegate used to validate that the CredentialID is unique to this user.</param>
+    /// <param name="requestTokenBindingId">DO NOT USE - Deprecated, but kept in code due to conformance testing tool</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> used to propagate notifications that the operation should be canceled.</param>
     /// <returns></returns>
     public async Task<RegisteredPublicKeyCredential> MakeNewCredentialAsync(
         AuthenticatorAttestationRawResponse attestationResponse,
         CredentialCreateOptions originalOptions,
         IsCredentialIdUniqueToUserAsyncDelegate isCredentialIdUniqueToUser,
+        byte[]? requestTokenBindingId = null,
```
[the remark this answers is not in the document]
abergs: That's an interesting alternative to get it less visible!

Src/Fido2/Fido2.cs
```diff
@@ -114,6 +117,7 @@ public async Task<VerifyAssertionResult> MakeAssertionAsync(
         IReadOnlyList<byte[]> storedDevicePublicKeys,
         uint storedSignatureCounter,
         IsUserHandleOwnerOfCredentialIdAsync isUserHandleOwnerOfCredentialIdCallback,
+        byte[]? requestTokenBindingId = null,
```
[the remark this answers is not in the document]
abergs: Also a good idea!

Src/Fido2/Extensions/CryptoUtils.cs
```diff
@@ -49,7 +49,7 @@ public static HashAlgorithmName HashAlgFromCOSEAlg(COSE.Algorithm alg)
         };
     }
 
-    public static bool ValidateTrustChain(X509Certificate2[] trustPath, X509Certificate2[] attestationRootCertificates)
+    public static bool ValidateTrustChain(X509Certificate2[] trustPath, X509Certificate2[] attestationRootCertificates, bool conformance = false)
```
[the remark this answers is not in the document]
abergs: Good idea!

Src/Fido2/Extensions/CryptoUtils.cs
abergs: @iamcarbon Not sure - I haven't looked into it.

Test/CryptoUtilsTests.cs
abergs: just a reminder to give these changes a review @aseigler :)

Src/Fido2/Extensions/CryptoUtils.cs
abergs: As part of #554 I reverted this change until we get more data on this or @aseigler can review. I will open another PR to bring these changes back.

Test/CryptoUtilsTests.cs
abergs: This was also reverted in #554

Src/Fido2/TrustAnchor.cs
googyi: https://github.com/passwordless-lib/fido2-net-lib/pull/456

TrustAnchor.cs : 32
This exact testcase was fixed with this (conformance tests v1.7.15):

Server-ServerAuthenticatorAttestationResponse-Resp-5 Test server processing "packed" FULL attestation F-10 Send ServerAuthenticatorAttestationResponse with FULL "packed" attestation, with attStmt.x5c containing full chain, and check that server returns an error https://datatracker.ietf.org/doc/html/rfc5280#section-6.1

Src/Fido2/TrustAnchor.cs
abergs: Resolved in: https://github.com/passwordless-lib/fido2-net-lib/pull/531/commits/dfa6f72b20d923fa917e48fbd7eeed24752d3921

Src/Fido2/Extensions/CryptoUtils.cs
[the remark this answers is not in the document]
abergs: Resolved in Resolved in: https://github.com/passwordless-lib/fido2-net-lib/pull/531/commits/dfa6f72b20d923fa917e48fbd7eeed24752d3921

Src/Fido2/Fido2.cs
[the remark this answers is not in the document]
abergs: @iamcarbon While drafting an implementation on your feedback  I decided against it since it would pollute the CredentialCreateOptions; they are expected to get serialised and sent to the client and should contain the "request token id" value.

Src/Fido2/Fido2.cs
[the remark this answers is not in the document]
abergs: Yes, let's do this in a separate PR.

Src/Fido2/TrustAnchor.cs
abergs: @googyi Thanks for chiming in. What part of the [datatracker.ietf.org/doc/html/rfc5280#section-6.1](https://datatracker.ietf.org/doc/html/rfc5280#section-6.1) is deciding checking the last entry of the trustpath to root certs?

googyi: I think this 6.1 section was referenced in that testcase, I don't know which part describes this exactly. 
This looks almost like that: 

"When the trust anchor is provided in the form of a self-signed
   certificate, this self-signed certificate is not included as part of
   the prospective certification path."

   But I am not sure that this is a self-signed cert.

abergs: 6.1 also mentions " A certificate MUST NOT appear more than once in a prospective
   certification path." which may be fixed by this?

PR #541, opened 2024-08-26
`FidoEnumConverter`: fall back to deserializing by name even if `EnumMemberAttribute` is set

Tyrrrz:
Currently, `FidoEnumConverter` works in the following way:

1. If an enum value is annotated by `EnumMemberAttribute`, then the provided custom name is used both for serialization and deserialization.
2. If the attribute is absent, then the default name (i.e. field name) is used.

This PR adds another fallback layer:

3. If an enum value is annotated by `EnumMemberAttribute`, but the default name is provided (not custom) then the deserialization would still succeed. Serialization will work the same regardless.

In other words, it would correctly resolve `PublicKeyCredentialType.PublicKey` from both `public-key`, as well as `PublicKey`.

This makes it a bit easier for applications to use Fido2 types directly in their payloads without building their own converters, and without relying on the intricacies of Fido2 naming conventions.

### DIFF
--- a/Src/Fido2.Models/Converters/FidoEnumConverter.cs
+++ b/Src/Fido2.Models/Converters/FidoEnumConverter.cs
@@ -12,23 +12,36 @@ public sealed class FidoEnumConverter<[DynamicallyAccessedMembers(DynamicallyAcc
         switch (reader.TokenType)
         {
             case JsonTokenType.String:
-                string text = reader.GetString();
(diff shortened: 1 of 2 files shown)


iamcarbon:
I had previously removed the relaxed parsing behaviors from this type, and believe they should remain strict. As a security critical library, we should continue to prioritize conformance with the specification 
over convenience.

PR #555, opened 2024-10-18
Conformance fix: Remove trustpath subject-issuer comparison

abergs:
This was originally written by @googyi (in #456), but due to lack of review in #531 we couldn't merge it.

@googyi Could you help explain this change?

[Notworthy comment](https://github.com/passwordless-lib/fido2-net-lib/pull/531#discussion_r1688633508) from @aseigler:

> The check to see if the subject and the issuer are the same is based on logic described in [datatracker.ietf.org/doc/html/rfc5280#section-3.2](https://datatracker.ietf.org/doc/html/rfc5280#section-3.2). I thought we had already covered this before, where the batch certificate referenced in the metadata is not a root...I'll have to review this more thoroughly.

### DIFF
--- a/Src/Fido2/Extensions/CryptoUtils.cs
+++ b/Src/Fido2/Extensions/CryptoUtils.cs
@@ -61,7 +61,7 @@ internal static class CryptoUtils
         // Let's check the simplest case first.  If subject and issuer are the same, and the attestation cert is in the list, that's all the validation we need
 
         // We have the same singular root cert in trustpath and it is in attestationRootCertificates
-        if (trustPath.Length == 1 && trustPath[0].Subject.Equals(trustPath[0].Issuer, StringComparison.Ordinal))
+        if (trustPath.Length == 1)
         {
             foreach (X509Certificate2 cert in attestationRootCertificates)
             {
--- a/Test/CryptoUtilsTests.cs
+++ b/Test/CryptoUtilsTests.cs
@@ -66,8 +66,8 @@ public class CryptoUtilsTests
 
         Assert.False(0 == attestationRootCertificates[0].Issuer.CompareTo(attestationRootCertificates[0].Subject));
         Assert.True(CryptoUtils.ValidateTrustChain(trustPath, attestationRootCertificates));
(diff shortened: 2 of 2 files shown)


abergs:
I've tried my best to iron this out, although I confess I'm not experienced in certs/trust path validation etc.

Looking at the comments (that are pasted from the linked MDS spec) - it seems the code is a bit too restrictive currently?

The relevant [part](https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-metadata-statement-v2.0-id-20180227.html#widl-MetadataStatement-attestationRootCertificates) from the MDS spec (also commented in the code already):

> attestationRootCertificates
> Each element of this array represents a PKIX [RFC5280] X.509 certificate that is a valid trust anchor for this authenticator model.
> Multiple certificates might be used for different batches of the same model.
> The array does not represent a certificate chain, but only the trust anchor of that chain.
> A trust anchor can be a root certificate, an intermediate CA certificate **or even the attestation certificate itself.**

Since this comment appears to be from the FIDO metadata specification, specifically around MetadataStatement.attestationRootCertificates, I believe it explicitly states that the trust anchor doesn't have to be self-signed (subject = issuer) - it can be an intermediate CA certificate or the attestation certificate itself, which would typically not have matching subject and issuer fields.

The thumbprints of the trustpath.0 and rootCert does line up, so it's only the first if that is questionably too restrictive.


Again, here's the code with what I think is an invalid comment:

```csharp

        // Let's check the simplest case first.  If subject and issuer are the same, and the attestation cert is in the list, that's all the validation we need

        // We have the same singular root cert in trustpath and it is in attestationRootCertificates
        if (trustPath.Length == 1 && trustPath[0].Subject.Equals(trustPath[0].Issuer, StringComparison.Ordinal))
        {
            foreach (X509Certificate2 cert in attestationRootCertificates)
            {
                if (cert.Thumbprint.Equals(trustPath[0].Thumbprint, StringComparison.Ordinal))
                {
                    return true;
                }
            }
        }
```

abergs:
@aseigler Does my analysis above make sense?

PR #556, opened 2024-10-18
Wrap arguments into classes

abergs:
Reason: In order to make arguments more readable and future API changes less likely to cause binary breaking changes to everyone we're moving many arguments into wrapping classes.

In this PR the classes are suffixed with Params, instead of popular `Options`/`Request` to avoid adding confusion as those terms are used elsewhere as part of webauthn.

PR #562, opened 2024-10-29
Introduce wrapping objects for Options-methods

### LINKED ISSUE #560: Wrap arguments to GetAssertionOptions and RequestNewCredential into Params-classes
This makes our API more consistent.

abergs:
This PR aligns with #556 and resolves #560.

In order to let the API surface grow (e.g. introducing Hints, allow RPs to select algo's to support, etc) we want to move towards this more flexible approach.

The main downside I see is that it's harder for devs to know which arguments are the main arguments without reading docs. Not sure how to best adress that, since most arguments used are not required.


abergs:
Will merge this in the next day, but I still welcome opinions if you have any @iamcarbon

Issue #566, opened 2024-10-30
Drop DevicePubkey-extension


abergs:
DPK has been removed from WebAuth L3:

* https://github.com/w3c/webauthn/pull/1957
* https://github.com/w3c/webauthn/pull/2109

PR #567, opened 2024-10-30
Drop support for device public key (dpk)

abergs:
- **Removed DevicePubKey p1**
- **Removed DevicePubkey p2**
- **Clean up crumbs**
- **format**


abergs:
Fixes #566. This PR removes support for device public keys as they have been dropped from the L3 spec.

It's a bit heartbreking to remove this much feature code, but if dpk's make it back as supplemental keys the implementation would probably look quite different.

abergs:
@aseigler Please let me know if you have any blocking thoughts.

aseigler:
It's disappointing that this didn't make it, but that's the way it goes.  Looks reasonable to me.

PR #575, opened 2024-11-08
Use BCL Base64Url implementation

iamcarbon:
This PR removes our custom Base64Url class, and takes a dependency on the new BCL implementation. I'll update this next week with the RTM package.

NOTE: our current implementation supported decoded Base64 content, which the new Base64Url implementation is strict, and throws when decoding non-base64 url chars. This PR also fixes all our test cases that were not properly encoding data.


iamcarbon:
@abergs Ready for review

joegoldman2 (approved):
LGTM

Tests/Fido2.Tests/TestFiles/attestationOptionsU2F.json
```diff
@@ -4,10 +4,10 @@
   "rp":{
     "name":"localhost",
     "id":"localhost"},
-  "user":{
-    "name":"aseigler",
-    "id":"ABC",
```
iamcarbon, line 9: NOTE: this is not a valid id, which should a valid Base64Url encoded byte array.

abergs:
> NOTE: our current implementation was not strict and also decoded Base64 content. The new implementation throws when encountering content that is not properly encoded.

I'm a bit worried about the rough edges this will cause as it will undoubtedly cause in-production bugs for some payloads. I'm wondering if we could implement a fallback that causes less problems for consumers?

iamcarbon:
I believe these "bugs" in production code should be fixed to be conformant with the WebAuthN specification. This is our last chance to fix this - and if we add a fallback, we may be allowing more clients to accidentally break compliance in the future.

Let me add some better error messaging in this PR for now. If you think we should continue to fallback to allow Base64 content to be decoded on Base64Url typed properties, we can add a fallback in our Base64UrlConverter.

We can also accept the break in 4.0, and add back the relaxed Base64 encoding behavior in 4.0.1, if this turns out to be a problem.

abergs:
> This is our last chance to fix this - and if we add a fallback, we may be allowing more clients to accidentally break compliance in the future.

> We can also accept the break in 4.0, and add back the relaxed Base64 encoding behavior in 4.0.1, if this turns out to be a problem.

I agree with and appreciate your concern, but I'm worried about teams/devs adapting to all of our breaking changes only to end up with the surprise that *some* passkeys are broken in production - which ultimately causes dissatisfaction with passkeys for both devs and end users. The amount of changes between v3 and v4 might make a simple rollback hard for many teams.

I suggest we implement a temporary fallback configuration option for 4.0 rather than releasing it in a 4.0.1 patch, if possible? 

Proposed approach:

* Add a simple config flag in 4.0 to enable the legacy relaxed encoding
* Teams can use this as an immediate fix if they encounter issues
* This buys them time to properly diagnose and fix the root cause
* We can remove this escape hatch in a future release once teams have migrated

This gives us the benefits of the breaking change while providing a safety net for teams who need it.

iamcarbon:
Agreed on the flag: 

How's this look:

```csharp
Base64UrlConverter.AllowRelaxedDecoding = true;
```

as an escape hatch to fallback to the old decoding behavior. Encoding was always strict.

Src/Fido2.Models/Converters/Base64UrlConverter.cs
```diff
@@ -8,20 +12,71 @@ namespace Fido2NetLib;
...
+        {
+            return Base64Url.DecodeFromUtf8(source);
+        }
+        catch
+        {
+            if (Base64.IsValid(source))
+            {
+                if (EnableRelaxedDecoding)
+                {
+                    return Base64Url.DecodeFromUtf8(source);
```
abergs: This should perhaps be `Base64.Decode...`?

iamcarbon: DUH. Let me also add a test case verifying this behavior.

abergs (approved):
Nicely done!

PR #579, opened 2024-11-18
Allow customised pubKeyCredParams

### LINKED ISSUE #533: Allow dev to customize what algo's they support instead of hard coding in CredentialCreateOptions
For conformance we must support RS1, but it should be left out from being enabled by default (suggested by @aseigler, could you expand the reasoning for disabling RS1?)

abergs:
This fixes #533.

### DIFF
--- a/Demo/TestController.cs
+++ b/Demo/TestController.cs
@@ -74,7 +74,9 @@ public class TestController : Controller
             ExcludeCredentials = existingKeys,
             AuthenticatorSelection = opts.AuthenticatorSelection,
             AttestationPreference = opts.Attestation,
-            Extensions = exts
+            Extensions = exts,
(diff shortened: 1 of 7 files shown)


abergs:
@aseigler Let me know if you think this looks good.

This allows devs to customise what algo's they want to support when creating the options. By default it's the same list as before, minus RS1.

In conformance testing we added RS1 to the list.

I've run it through conformance and it passes.

aseigler:
> @aseigler Let me know if you think this looks good.
> 
> This allows devs to customise what algo's they want to support when creating the options. By default it's the same list as before, minus RS1.
> 
> In conformance testing we added RS1 to the list.
> 
> I've run it through conformance and it passes.

We had talked about this before, looks good to me!

PR #581, opened 2024-11-22
[Models] Enable nullable (1 of x)

### LINKED ISSUE #466: Enable nullable for Fido2.Models
I would be really helpful to annotate the nullability of the `Fido2.Models` types. In other words, to set `<Nullable>enable</Nullable>` in the `Fido2.Models` project.

This way, it would become clear for FIDO2 newbies to understand which information is optional and which one is required.

For example, in the demo, you have a line of code that stores a credential coming from `MakeNewCredentialAsync`:

    DevicePublicKeys = new List<byte[]>() { success.Result.DevicePublicKey }

If both `Fido2.Models` and `Demo` were NRT enabled, this would pop up as a compilation error. Righteously so, because I don't think it make sense to store a list containing a `null` along the credential. So in this example, it would really help if `RegisteredPublicKeyCredential.DevicePubliKey` was declared as `byte[]?`. And I assume there's plenty more.

iamcarbon:
This is the first of a few PR's to enable nullable on the models project and close #466 

This PR:

- Matches the requiredness of various properties with the specifications. Some properties were annotated with required instead, to limit churn in our tests
- Makes a few more classes immutable

The next PR will turn enable nullable at the project level

Posting for early review / feedback.


iamcarbon:
@joegoldman2 @abergs Review for review / feedback. We should get these final nullability changes in before the 4.0 release.

iamcarbon:
@abergs Quick ping on this. There's one more PR after this to flipping on <Nullable>enable</Nullable> on the project and updating the directives. That should be the last of the churn for Models in the 4.0 cycle.

abergs (approved):
Thank you @iamcarbon, looks good!

PR #586, opened 2024-12-30
Base64Url Id instead of byte[]

Regenhardt:
Also after a new install, dotnet wouldn't accept some packages, so I changed the nuget.config according to [the nuget signing blog post](https://devblogs.microsoft.com/nuget/the-nuget-org-repository-signing-certificate-will-be-updated-as-soon-as-april-8th-2024/#allow-the-new-nuget-org-certificate). The post does not mention the certificate I removed with the commit, but I'll put it back if it's otherwise needed.


abergs (changes requested):
I've been reviewing this, as well has reading some of the related WebAuthn discussions. like: https://github.com/w3c/webauthn/issues/2119

Although this will introduce lots of "won't-compile"-errors when people upgrade to 4.0, I agree it's probably the right thing to do.

The current set of changes look good, however I think we also need to change the `Id`-prop on `AuthenticatorAttestationRawResponse`, so I will request changes.

Regenhardt:
Changed the `Id` of `AuthenticatorAttestationRawResponse` to string too.

This might also be the right time to make the `RawId` be sent raw according to the comment in `AuthenticatorAssertionRawResponse`, although b64url is indeed a way nicer and also a widely expected way to send that kind of data:

```csharp
// might be wrong to base64url encode this...
[JsonConverter(typeof(Base64UrlConverter))]
[JsonPropertyName("rawId"), Required]
public byte[] RawId { get; init; }
```

Regenhardt:
Looks like for some reason the upload artifats github action is deprecated. Should I just update it here or would your prefer it in its own branch?

abergs:
@Regenhardt It's own branch please!

abergs:
Is this ready for review again?

Regenhardt:
Yes it's ready for review, although the failed pipeline has to wait for the new artifacts action to be updated.

Regenhardt:
This needs #590 to fix the pipeline.

PR #589, opened 2025-02-05
Use JSON source generator & enable trimming in DistributedCacheMetadataService

eduar-hte:
This is an update to be able to use `DistributedCacheMetadataService` in an AOT build, based on work in PR #304.

In order to store and retrieve the cached metadata blob, calls to `JsonSerializer`'s `Serialize` & `Deserialize` need to specify `FidoModelSerializerContext.Default.MetadataBLOBPayload`.


eduar-hte:
@dotnet-policy-service agree

eduar-hte:
@dotnet-policy-service agree

PR #601, opened 2025-08-29
Enable packing for Fido2.BlazorWebAssembly project

Regenhardt:
Turns out, release 4.0.0 didn't include the WebAssembly package. Whoops.

### DIFF
--- a/Src/Fido2.BlazorWebAssembly/Fido2.BlazorWebAssembly.csproj
+++ b/Src/Fido2.BlazorWebAssembly/Fido2.BlazorWebAssembly.csproj
@@ -7,6 +7,7 @@
     <Nullable>enable</Nullable>
     <IsTrimmable>true</IsTrimmable>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <IsPackable>true</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


Regenhardt:
@abergs It would be nice to have this package as a 4.0.0 release also, is that possible?

abergs:
Hey @Regenhardt. I'm not a keen Blazor developer, I'm debating if I really want to take the responsibility of owning this package on or not.

I'm assuming this package will need updates as Blazor patterms evolve (It's relatively young as a framework) and I don't think I will be able to support that workload (I'm not an avid Blazor developer).

I think perhaps it's better if someone else owns and maintains that package?

abergs:
On second thought; the source of this package is Tiny and may not need to evolve that much over time. If you agree that there will likely not be a lot of changes, I'm ok with publishing it.

Regenhardt:
I expect there won't be too many more architectural changes to Blazor/JS-Interop now. I'll gladly keep this in mind if I notice any bigger changes in Blazor to check if it's still working. Also I expect a few people to actually use it so there's a good chance someone will notice and send a PR in case anything does indeed break it. Thanks for adding it, I'm glad will be under the same umbrella as the main packages.

Issue #550, opened 2024-10-06
[API Proposal] Allow to disable caching for Metadata or use another cache mechanism 


ghost:
## Context

Today there is no easy way to use another cache than `IDistributedCache` (like the new HybridCache of .NET 9, FusionCache, etc) when using the metadata service. The `AddCachedMetadataService` method is the entry point for register the metadata service, registering a service that is based on `IDistributedCache` and `IMemoryCache`.

In addition, there is no way to completely disable the cache as it is directly implemented in the only `IMetadataService` implementation available out of the box (`DistributedCacheMetadataService`).

To solve this problem, it would be nice to have the possibility to register the service and repositories without any caching, which leaves the user the possibility to add the cache afterwards. Of course it will still be possible to register a default cache implementation that is based, as today, on `IDistributedCache` and `IMemoryCache`.

## API Proposal

I propose the following API:

```diff
namespace Fido2NetLib;

+    public sealed class DefaultMetadataService : IMetadataService // An implementation without any cache mechanism
```

```diff
namespace Microsoft.Extensions.DependencyInjection;

public static class Fido2NetLibBuilderExtensions
{
-    public static void AddCachedMetadataService(this IFido2NetLibBuilder builder, Action<IFido2MetadataServiceBuilder> configAction);
     // Will register the default service without any cache
+    public static void AddMetadataService(this IFido2NetLibBuilder builder, Action<IFido2MetadataServiceBuilder> configure);
     // As the cache is currently implemented on service level (should be probably more flexible on repository level + decorator pattern but it's outside the scope of this issue),
     // this method will replace the current registered implementation by DistributedCacheMetadataService.
+    public static IFido2MetadataServiceBuilder UseDistributedCache(this IFido2MetadataServiceBuilder builder);
}
```

I am open to suggestions for the name of `UseDistributedCache` method for which I am not really inspired.

## API Usage

```cs
var services = new ServiceCollection();
services.AddFido2(options => { ... }).AddMetadataService(options =>
{
    options.AddFidoMetadataRepository();
    options.UseDistributedCache();
});
```

instead of that, currently:

```cs
var services = new ServiceCollection();
services.AddFido2(options => { ... }).AddCachedMetadataService(options =>
{
    options.AddFidoMetadataRepository();
});
```

## More context

#397 also reports some limitations of the current model

cc @abergs @iamcarbon

joegoldman2:
I would be happy to help with the implementation if you agree with the proposal.

[...]

Issue #397, opened 2023-06-05
MetadataService/DistributedCacheMetadataService IDistributedCache dependency limitations


Simonl9l:
Great well written and (for the most part) easy to use library - thanks!

The `DistributedCacheMetadataService` leverages both `IMemoryCache` and `IDistributedCache`. However it makes assumptions on how much data can be stored in any givrn Distributed Cache Implementation.

In the case of using say [this](https://github.com/awslabs/aws-dotnet-distributed-cache-provider) AWS implementation that uses DynamoDB as its backing store. Here the max size that can be persisted is 65535 bytes. It seems that the Metadata blob is larger than this and fails.

In this case it's a physical limitation of the DynamoDB that can not be overcome. Arguably the AWS `IDistributedCache` implementation should perhaps handle it (with perhaps multiple entries), but there is a chain of assumptions here that might lead to opinionated code.
 
Whilst it would be good to get a solution to this, in the meantime per the docs in this area it is not clear if there is an alternative to _**not**_ using using the `DistributedCacheMetadataService` as defined by `.AddCachedMetadataService` and be just reliant on the local file system to cache the metadata file?

It seem that `AddFileSystemMetadataRepository` can only be using with the context of the  `.AddCachedMetadataService` to configure it but this then brings in the `DistributedCacheMetadataService` that in turn brings in the `IMemoryCache` and more specifically the `IDistributedCache` that is otherwise used for session management that can be used easily within the 65535 byte limitation it has.

abergs:
Please see #620. Closing this issue in order to move the discussion to that PR. 

Please note that I think this is already achievable using DI in the 4.0.0 release.

PR #620, opened 2025-09-03
Simplify the MDS Builder API

abergs:
This PR aims to simplify the MDS builder API and adress some current problems we're seeing.

Developers today seems to believe it's hard / unfeasible to implement their own "caching" layer for MDS data.

* #550
* #397 

This is because we have not documented the current architecture properly, and the current builder pattern does indicate that there is not a lot of flexibility in the design, although the opposite is true.

This PR adds documentation, as well as reshapes the API to better surface the flexibility.

## Example in the API change

**Before**
```csharp
services
    .AddFido2(...)
    .AddCachedMetadataService(config =>
    {
        config.AddFidoMetadataRepository(httpClientBuilder =>
        {
            //TODO: any specific config you want for accessing the MDS
        });
    });
```

**After:**

```csharp
  services
      .AddFido2(...)
      .AddFidoMetadataRepository()
      .AddCachedMetadataService();
```

We're also adding a two new builder methods, that makes it clear that you can register your own MDS Service  to wrap the repositories, as well as custom repositories

**Using FIDO mds repository + your own custom service to do caching, logging, storage, whatever**

```csharp
services
      .AddFido2(...)
      .AddFidoMetadataRepository()
+     .AddMetadataService<MyCustomMetadataService>();
```

**Using a completely bespoke solution to fetch mds data and access it**

```csharp
services
      .AddFido2(...)
+     .AddMetadataRepository<MyCustomMetadataRepository>()
+     .AddMetadataService<MyCustomMetadataService>();
```

## Breaking changes

This PR introduces some breaking changes in the API builder (compile time) and some run time breaking changes, namely: Removes the default registration of the NullMetadataService

* Remove the `IFido2MetadataServiceBuilder `
* Removes the default registration `NullMetadataService`

We are also adding plenty of comments as well as hiding the ConformanceMDS from Intellisense. I wonder if we should really remove it instead?

### DIFF
--- a/Tests/Fido2.AspNet.Tests/AddFido2ExtensionTests.cs
+++ b/Tests/Fido2.AspNet.Tests/AddFido2ExtensionTests.cs
@@ -95,4 +95,142 @@ public class AddFido2ExtensionTests
         // var mds = serviceProvider.GetService<IMetadataService>();
         // Assert.Null(mds);
     }
+
+    [Fact]
+    public void AddMetadataService_RegistersCustomMetadataService()
+    {
+        // Arrange
+        var services = new ServiceCollection();
+        var builder = services.AddFido2(config => { });
+
+        // Act
+        builder.AddMetadataService<TestMetadataService>();
+
+        // Assert
+        var serviceProvider = services.BuildServiceProvider();
+        var metadataService = serviceProvider.GetService<IMetadataService>();
+        Assert.NotNull(metadataService);
+        Assert.IsType<TestMetadataService>(metadataService);
+    }
+
+    [Fact]
+    public void AddCachedMetadataService_RegistersCachedService()
+    {
+        // Arrange
+        var services = new ServiceCollection();
+        services.AddLogging();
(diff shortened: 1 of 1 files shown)


abergs:
Please note, this PR does not actually add any new feature to the project. It does however help clarify how existing features can solve a users problem.

Does this suit your need @Simonl9l, @joegoldman2. Also curious if you have any thoughts @iamcarbon and @aseigler.

I've left it out of this PR, but I am interested in writing a `FileSystemBlobRepository` that sources the mds.jwt blob from the file system instead of multiple json files that the current `FileSystemRepository` and a "uncached" service that sits on top of it.

joegoldman2:
It's just weird for me that the cache is at the service level and not at the store level (with a decorator for example).

abergs:
@joegoldman2 When you say store level, you mean the repository?

joegoldman2:
> @joegoldman2 When you say store level, you mean the repository?

Yes, the repository sorry.

PR #642, opened 2025-12-09
Handle malformed assertion signatures

explunit:
So far this PR is only the unit tests that illustrate what (I think) is a problem, and I am interested in feedback before I attempt the fix.

My expectation was that any type of invalid signature, regardless of how malformed it was, would throw the `Fido2VerificationException` with message "Signature does not match". But, as illustrated by these unit tests, some signatures throw other exceptions:

```
Assert.Throws() Failure: Exception type was not an exact match
Expected: typeof(Fido2NetLib.Fido2VerificationException)
Actual:   typeof(System.Formats.Asn1.AsnContentException)
---- System.Formats.Asn1.AsnContentException : The ASN.1 value is invalid.
```

and 

```
 Assert.Throws() Failure: Exception type was not an exact match
 Expected: typeof(Fido2NetLib.Fido2VerificationException)
 Actual:   typeof(System.ArgumentOutOfRangeException)
 ---- System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
```

Should the `VerifyAsync` method trap these exceptions or should the underlying `CryptoUtils.SigFromEcDsaSig` throw a `Fido2VerificationException`? I see that another method in that file, `HashAlgFromCOSEAlg`, _does_ throw a `Fido2VerificationException`.

### DIFF
--- a/Tests/Fido2.Tests/AuthenticatorResponse.cs
+++ b/Tests/Fido2.Tests/AuthenticatorResponse.cs
@@ -2461,8 +2461,12 @@ public class AuthenticatorResponseTests
         Assert.Equal(Fido2ErrorMessages.MissingStoredPublicKey, ex.Message);
     }
 
-    [Fact]
-    public async Task TestAuthenticatorAssertionInvalidSignature()
+    [Theory]
+    [InlineData(new byte[] { 0x30, 0x00 }, true)]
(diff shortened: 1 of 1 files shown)


abergs:
Implemented in #652

PR #654, opened 2026-04-12
Demo: Use PublicKeyCredential parse functions instead of converting by hand

Edduaardo:
Use `PublicKeyCredential` functions `parseCreationOptionsFromJSON()`, `parseRequestOptionsFromJSON()` and `toJSON()` to convert between Base64Url and ArrayBuffers instead of doing it by hand, as it is the recommended way now. It also stops using the deprecated `extensions` property in favor of `clientExtensionResults`.

https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/parseCreationOptionsFromJSON_static
https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/parseRequestOptionsFromJSON_static
https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/toJSON


Edduaardo:
@dotnet-policy-service agree

abergs (approved):

abergs:
Thank you for the contribution @Edduaardo!

PR #655, opened 2026-04-15
Restore EnumMemberAttribute for .net10

vpetrusevici:
#### Summary
This PR updates the attribute pattern used for enum members across the project to ensure full compatibility with both `System.Text.Json` (specifically in .NET 9+) and `Newtonsoft.Json`.

#### Problem
Previously, enum members used a conditional `#if NET9_0_OR_GREATER` block with an `#else` clause that toggled between `[JsonStringEnumMemberName]` and `[EnumMember]`. 
```csharp
#if NET9_0_OR_GREATER
    [JsonStringEnumMemberName("value")]
#else
    [EnumMember(Value = "value")]
#endif
```
While this worked for `System.Text.Json`, it broke compatibility with `Newtonsoft.Json` (and tools like **NSwag** that rely on it) when running on .NET 9+. Since `Newtonsoft.Json` does not recognize `[JsonStringEnumMemberName]`, it would fall back to default naming, leading to serialization mismatches in environments where `System.Text.Json` is not the primary serializer.

#### Solution
The `#else` block has been removed, making `[EnumMember]` unconditional. `[JsonStringEnumMemberName]` remains wrapped in `#if NET9_0_OR_GREATER` to take advantage of native .NET 9 features while preserving the metadata needed by other libraries.

**New Pattern:**
```csharp
#if NET9_0_OR_GREATER
    [JsonStringEnumMemberName("public-key")]
#endif
    [EnumMember(Value = "public-key")]
    PublicKey,
```

#### Key Changes
- Removed `#else` from enum member attributes in all model files.
- Ensured `[EnumMember]` is always present for every enum member.
- Maintained existing `#if` logic for `JsonConverter` declarations and internal logic where appropriate.
- Verified that these changes preserve correct serialization behavior in NSwag and other Newtonsoft.Json-dependent consumers.

### DIFF
--- a/Src/Fido2.Models/Fido2Configuration.cs
+++ b/Src/Fido2.Models/Fido2Configuration.cs
@@ -126,6 +126,11 @@ public class Fido2Configuration
     /// </summary>
     public CredentialBackupPolicy BackedUpCredentialPolicy { get; set; } = CredentialBackupPolicy.Allowed;
 
+#if NET9_0_OR_GREATER
+    [JsonConverter(typeof(JsonStringEnumConverter<CredentialBackupPolicy>))]
+#else
+    [JsonConverter(typeof(FidoEnumConverter<CredentialBackupPolicy>))]
+#endif
     public enum CredentialBackupPolicy
     {
         /// <summary>
--- a/Src/Fido2.Models/Fido2Configuration.cs
+++ b/Src/Fido2.Models/Fido2Configuration.cs
@@ -133,9 +133,8 @@ public class Fido2Configuration
         /// </summary>
 #if NET9_0_OR_GREATER
         [JsonStringEnumMemberName("required")]
-#else
-        [EnumMember(Value = "required")]
 #endif
+        [EnumMember(Value = "required")]
         Required,
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


vpetrusevici:
@abergs sorry, we need to restore this attribute :)
It's still needs for Newtonsoft.Json cases...

abergs:
@vpetrusevici Have we verified that the original issue is still resolved when EnumMember is added?

And, is there a way to add tests to confirm the proper behavior in serialization?

vpetrusevici:
Yes. I'll try to add test, including Newtonsoft

vpetrusevici:
> @vpetrusevici Have we verified that the original issue is still resolved when EnumMember is added?
> 
> And, is there a way to add tests to confirm the proper behavior in serialization?

tests added

vpetrusevici:
@abergs anything else? :)

abergs:
All good, preview.3 will be rolling out.

vpetrusevici:
@abergs hi! did you release it?

vpetrusevici:
@abergs, preview.3 is still unavailable :(

abergs:
@vpetrusevici Seems there was a github issue while running tests. Running everything again.

